PR #117, opened 2020-12-01
Speedup scanning workspace directory; Support EXCLUDED_DIRECTORIES config parameter

kolomenkin:
## Proposed Changes

1. Previous scan complexity for checking for exclusions was: M * N * K where
  M - number of folders inside of workspace directory
  N - average directory nesting level + 2 for `/tmp/lint`
  K - number of elements in excluded directory list; By default it has length 8

New scan complexity is linear to M.

2. Do not walk into excluded directories. My current project has 20K files inside of excluded directories (virtual environments).
  Scanning all of them takes a lot of time (15+ seconds). New implementation scans in less than a second.

3. Added possibility to customize excluded directory list via EXCLUDED_DIRECTORIES configuration parameter.
  My current project contains second virtual environment and also `.temp` directory with numerous files outside of git source control.
  So I can skip scanning thouthands of files in hundreds of directories easily now.

### Author/Contributor

### Reviewing Maintainer
- [`performance`, `enhancement`] Label as either `automation`, `bug`, `documentation`, `enhancement`, `infrastructure`, or `performance`


kolomenkin:
I do not get why do you have two change logs. I updated both of them.

Feel free to request any changes.

I will fix merge conflicts between my PRs once one of them is merged. Don't worry.

P.S. Can you tell when approximately it would be possible to get tagged docker image with performance fix?

kolomenkin:
By the way. Skipping directories during `walk` is based on this code sample:

https://stackoverflow.com/a/19859907/739731

nvuillam:
> I do not get why do you have two change logs. I updated both of them.
> 
> Feel free to request any changes.
> 
> I will fix merge conflicts between my PRs once one of them is merged. Don't worry.
> 
> P.S. Can you tell when approximately it would be possible to get tagged docker image with performance fix?

Once a PR is merged in master, there is Deploy-Prod job triggered. "insiders" version ( docker image tag: latest ) is then available about 30 mn after
I can also generate a new official patch anytime (but I prefer to test with insiders version before), I can probably do it tonight if you're in a hurry :)

nvuillam (changes requested):
Thanks for the update, I'm still a beginner in python (I started with Mega-Linter) so any proposition, especially about performances, is gladly welcome :)
About EXCLUDED_DIRECTORIES, it was already possible to do that with FILTER_REGEX_EXCLUDE, but I agree that when you have several directories, it can quickly become unreadable, so I agree with your proposal.

But i have some questions /remarks:
- Maybe using a "glob" format would be better ? ( like `folder_x/folder_y/**` )
for example, if you have a repo
```
- folder_x/folder_y
- folder_y
```
is there a way to ignore the second one but not the first one with current configuration capablilities ?

2. Test method for EXCLUDED_DIRECTORIES ? :)

+ code review comments directly in "updated files" tab

docs/configuration.md
```diff
@@ -66,6 +66,7 @@ report/
 | **DEFAULT_BRANCH**                | `master`              | The name of the repository default branch. Warning: In new github repositories, master branch is named `main`, so you need to override this value with `main`                    |
 | **DEFAULT_WORKSPACE**             | `/tmp/lint`           | The location containing files to lint if you are running locally.                                                                                                                |
 | **DISABLE_ERRORS**                | `false`               | Flag to have the linter complete with exit code 0 even if errors were detected.                                                                                                  |
+| **EXCLUDED_DIRECTORIES**          | \[...\]               | Customize excluded directory list. Default list of directories is stored in `megalinter/utils.py` in `list_excluded_directories` function.                                       |
```
nvuillam, line 69: You can describe it as "Additional excluded directories" ? It would be more an append to the hardcoded ones than a replacement

megalinter/MegaLinter.py
```diff
@@ -428,15 +428,9 @@ def list_files_all(self):
...
-        excluded_directories = utils.list_excluded_directories()
-        for (dirpath, _dirnames, filenames) in os.walk(self.workspace):
-            exclude = False
-            for dir1 in dirpath.split(os.path.sep):
-                if dir1 in excluded_directories:
-                    exclude = True
-                    break
-            if exclude is True:
-                continue
+        excluded_directories = set(config.get("EXCLUDED_DIRECTORIES", utils.list_excluded_directories()))
```
nvuillam, line 431: I really would prefer an append + unique of the lists, nobody will ever lint node_modules or .rbenv folders, so if they want to ignore additional folders, they should not have to redefine the defaults ones in configuration

nvuillam, line 431: use config.get_list , so it can manage variables from ENV and from .mega-linter.yml

nvuillam: deal :)

docs/configuration.md
kolomenkin: Maybe I should rename parameter in this case? ADDITIONAL_EXCLUDED_DIRECTORIES ?

nvuillam: nop it's ok you can let it like that
But i'd like you to remove the reference to the method... a user should not have to read the sources to know how a parameter works ^^
I'll update build.py to dynamically generate the hardcoded list of excluded directories in the documentation

megalinter/MegaLinter.py
```diff
@@ -428,15 +428,9 @@ def list_files_all(self):
...
-        excluded_directories = utils.list_excluded_directories()
-        for (dirpath, _dirnames, filenames) in os.walk(self.workspace):
-            exclude = False
-            for dir1 in dirpath.split(os.path.sep):
-                if dir1 in excluded_directories:
-                    exclude = True
-                    break
-            if exclude is True:
-                continue
+        excluded_directories = set(config.get_list("EXCLUDED_DIRECTORIES", utils.list_excluded_directories()))
```
nvuillam, line 431: @kolomenkin could you do something like below, so it's an apprend and not a replace ?

```python
set(config.get_list("EXCLUDED_DIRECTORIES",[]) + utils.list_excluded_directories())
```

kolomenkin:
> Thanks for the update, I'm still a beginner in python (I started with Mega-Linter) so any proposition, especially about performances, is gladly welcome :)
> About EXCLUDED_DIRECTORIES, it was already possible to do that with FILTER_REGEX_EXCLUDE, but I agree that when you have several directories, it can quickly become unreadable, so I agree with your proposal.
> 
> But i have some questions /remarks:
> 
> * Maybe using a "glob" format would be better ? ( like `folder_x/folder_y/**` )
>   for example, if you have a repo
> 
> ```
> - folder_x/folder_y
> - folder_y
> ```
> 
> is there a way to ignore the second one but not the first one with current configuration capablilities ?
> 
> 1. Test method for EXCLUDED_DIRECTORIES ? :)
> 
> * code review comments directly in "updated files" tab

Don't mind :)
Original super linter is even much slower for my project. With similar problems. And changing their bash implementation to be more clever and flexible is more complicated than changing python. I do not get why they ever used bash. They have 4+ GB docker image. They could use any language or technology.

Implementing `folder_x/folder_y/**` is more complicated. I would say current solution may help to ignore most problematic dirs by their short names. And more precise exclusions could be done via regex.
But I agree. This is a bit confusing we have two types of ignore lists.

From one hand we can leave as-is. This is not the worth solution :)
For example setup the list of short folder names is much easier than escaping complicated regex to be passed though bash. And in lots of cases this simplified exclusion list may be enough.

Alternatively I can propose the following solution:
1. Convert existing 8 dir names into one big regex. And set it as default value for exclude regex.
2. I will rewrite the code to match regex inside of the loop gathering file list. Once any directory matches exclusion regex I will not go inside. No matter if regex is created to allow internal items (for example regex: `/dir/subdir/$` ). And similar logic for positive regex.

As a side effect we will loose original number of files on disk found.

If anybody will want to redefine regex he should also correctly match dirs like `.git` by his own.

Optionally we could support list of regular expressions instead of single regex. But:
- it would be less efficient to match list of regexes in a loop. But we could try combining them in runtime.
- comma will becode unusable inside of regex because it will become a delimiter in environment var


**Please make some choice**

megalinter/MegaLinter.py
kolomenkin: Originally I wanted to change `list_excluded_directories`. The only one thing stopped me. It did not depend on `config` before. So I decided not to add this dependency. But this is a very minor argument in this case.

I will update `list_excluded_directories`. It will encapsulate calculation of this list better. And it will return a `set` from now, not a `list`

nvuillam:
> And changing their bash implementation to be more clever and flexible is more complicated than changing python. I do not get why they ever used bash. They have 4+ GB docker image. They could use any language or technology.

I originally implemented the regex filters on super-linter, + other stuff ... and when I saw that doing that in bash was such a pain in the ***, I proposed a big PR... that after discussion they refused with "we started in bash we continue in bash" explanation ... and mega-linter was born :D ( i did not initially intended to do a hard-fork of a tool made by github ... but I think the [advantages of modern language programming speak for themselves](https://nvuillam.github.io/mega-linter/mega-linter-vs-super-linter/) ! )
What is nice is that I still remain up to date with their code base, so a merge of Mega-Linter into Super-Linter could theoritecally be done in the future
Meanwhile, at each linter they add it takes one hour for me to add it in Mega-Linter thanks to YML files and automation 😄 

> **Please make some choice**
> From one hand we can leave as-is. This is not the worth solution :)

Let's go small first before going big : leave as-is is my choice , we'll see later if someone has specific needs not covered by current capabilities :)

kolomenkin:
> > And changing their bash implementation to be more clever and flexible is more complicated than changing python. I do not get why they ever used bash. They have 4+ GB docker image. They could use any language or technology.
> 
> I originally implemented the regex filters on super-linter, + other stuff ... and when I saw that doing that in bash was such a pain in the ***, I proposed a big PR... that after discussion they refused with "we started in bash we continue in bash" explanation ... and mega-linter was born :D ( i did not initially intended to do a hard-fork of a tool made by github ... but I think the [advantages of modern language programming speak for themselves](https://nvuillam.github.io/mega-linter/mega-linter-vs-super-linter/) ! )
> What is nice is that I still remain up to date with their code base, so a merge of Mega-Linter into Super-Linter could theoritecally be done in the future
> Meanwhile, at each linter they add it takes one hour for me to add it in Mega-Linter thanks to YML files and automation 😄
> 
> > **Please make some choice**
> > From one hand we can leave as-is. This is not the worth solution :)
> 
> Let's go small first before going big : leave as-is is my choice , we'll see later if someone has specific needs not covered by current capabilities :)

Maybe after some time it would be a good idea to propose them Python again.
They will be able to see how their code could look like. And take existing code into their repo.
Maybe they could support both implementations for some short time.

I feel like trying to overcome GitHubs solution may be inefficient in long term.

And it would be nice if Python solution would be more flexible and faster. I know it is not easy to flexible scan folders in bash. Implementation can easily fall into generating hundreds and thousands of child processes loosing performance.

By the way.
How do you update 3rd party linters? Just fetch latest versions all the time? No version lock or control?

nvuillam:
> By the way.
How do you update 3rd party linters? Just fetch latest versions all the time? No version lock or control?

I quickly got borded of dependabot PRs everyday, so I implemented the following behavior:
- everyday at midnight + at each PR, a GH job [Auto-update linters](https://github.com/nvuillam/mega-linter/actions?query=workflow%3A%22Auto-Update+Linters%22)
  - Build new image from Dockerfile
  - Call --help and --version on each linter (and store them in .automation/generated/linter-versions.json and linter-helps.json)
  - Build project ( build.sh)

If there are differences, create a PR, that I validate manually (automerge was supposed to merge them automatically if test cases pass, but it's not working yet, that's why i disabled it)

I see that push commits on repo forks does not work ( github PAT is required, and the only one stored is for this main repo), i'll try to see how i can deactivate that for PR from forks.
Meanwhile, you can [manually extract fixed sources in your branch](https://nvuillam.github.io/mega-linter/reporters/UpdatedSourcesReporter/) :)

kolomenkin:
> Test method for EXCLUDED_DIRECTORIES ? :)

Did I understand correctly you want unit test to be added for EXCLUDED_DIRECTORIES ?

I can test method `get_excluded_directories`

But it could a good idea to have tests for main functionality of `list_files_all`, `collect_files`.
With pre-creating some directories in runtime and scanning them.
Maybe you already have?

nvuillam:
> Did I understand correctly you want unit test to be added for EXCLUDED_DIRECTORIES ?
I can add test for function get_excluded_directories
But it could a good idea to have tests for main functionality of list_files_all, collect_files.
With pre-creating some directories in runtime and scanning them.
Maybe you already have these tests?

A test using EXCLUDED_DIRECTORIES will be enough

I already got a [pretty good code coverage](https://codecov.io/gh/nvuillam/mega-linter/src/master/megalinter/MegaLinter.py), tests may be not exhaustive, but all configurations are tested in other tests, for example FILTER_REGEX_INCLUDE is used to test success and failures of each linter

You can create anything you want under .automation/test , it's not packaged so it does not makes the Docker image heavier (but maybe in another PR so we can finalize this one ? )

nvuillam:
I see you have a [cspell failure](https://nvuillam.github.io/mega-linter/descriptors/spell_cspell/) in Mega-Linter: 

image

nvuillam:
@kolomenkin i can look away for the test class for now as it should not break existing uses, but plz update .cspell.json and copy-paste updated-sources (there is a fix from python isort :) )

image

image

kolomenkin:
> @kolomenkin i can look away for the test class for now as it should not break existing uses, but plz update .cspell.json and copy-paste updated-sources (there is a fix from python isort :) )
> 
> image
> 
> image

Thanks for screenshots! I tried to find those artifacts few times without looking screenshots and failed.

nvuillam:
There are some troubles with jobs when coming from PR, it seems Comments with summary and links, and Github Status are not displayed... I made some (uncompleted but saving crashes) updates, you may get up to date with master :)

If you have new ideas of PR for later, would you like an access as collaborator directly on the repo ?

kolomenkin:
> There are some troubles with jobs when coming from PR, it seems [Comments with summary and links](https://nvuillam.github.io/mega-linter/reporters/GitHubCommentReporter/), and [Github Status](https://nvuillam.github.io/mega-linter/reporters/GitHubStatusReporter/) are not displayed because of some github credential access issues)
> I made some (uncompleted but saving crashes) updates, you may get up to date with master :)
> 
> If you have new ideas of PR for later, would you like an access as collaborator directly on the repo ?

Thanks a lot.
I don't think being an official collaborator will help me somehow. I will need code review in any case.

nvuillam:
Merged ! there is still a typo in .cspell but i fix it in master.

You'll be able to use new insiders version in about half an hour, when step "Deploy docker image on dockerhub" in [this job](https://github.com/nvuillam/mega-linter/runs/1483479211?check_suite_focus=true) will be complete :)

- using @insiders instead of @v4 if you use Mega-Linter as a GitHub Action
- using mega-linter-runner@insiders if you use mega-linter-runner
- using docker image docker://nvuillam/mega-linter:v4

Thanks again for your contribution :)

nvuillam:
@kolomenkin i invited you as collaborator
I added validation rules on master so for the moment only I can merge PRs in master, but:
- Developing in internal branches will allow you to have all reporters working
- You can generate anytime a Mega-Linter **alpha version**, by manually **merging your devs in alpha branch** :)

PR #123, opened 2020-12-02
Optimize docker image layers for speed and reuse

kolomenkin:
Fixes #

## Proposed Changes

1. Reduce number of image layers when it does not hurt readability
2. Move copying local repo files into image as late as possible to reuse as much layers as possible when something is changed in local code.

But I understand it is not a critical change. I feel it will mostly help mega linter developers during local image rebuild.

Once you change some python code, it will take just few seconds to prepare updated image.

### Author/Contributor

### Reviewing Maintainer
- [`performance`] Label as either `automation`, `bug`, `documentation`, `enhancement`, `infrastructure`, or `performance`


nvuillam:
I merged into alpha so we'll be able to compare with current images :)

nvuillam:
This PR looks great, i think that associated with https://github.com/nvuillam/mega-linter/pull/121 (where i try to use packages instead of copying from docker images), it could really decrease the number of steps !
There is currently an issue with terrascan (there is a new violation that I don't succeed to avoid with example file, as i know nothing about terrascan), once solved I'll merge in master, then you'll be able to make your branch up to date with master, and all should be ok :)

nvuillam:
I merged the  terrascan test fix in master, if you refresh your branch with it, CI should pass 👍 
You may have a particular attention on conflicts in Dockerfile , sorry !

nvuillam:
Great, with both our PRs, we got from 93 to 83 layers for the image, good job ! 👍

PR #139, opened 2020-12-09
Customize linter file names and file extensions

kolomenkin:
Fixes #
- speed optimization by using sets, pre-compiled regular expressions

## Proposed Changes

- support regular expressions in file names
- add config setting <LINTER_NAME>_ADDITIONAL_FILE_EXTENSIONS
- add config setting <LINTER_NAME>_ADDITIONAL_FILE_NAMES_REGEX

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/megalinter/Linter.py
+++ b/megalinter/Linter.py
@@ -8,8 +8,8 @@ The following list of items can/must be overridden on custom linter local class:
 - field linter_url (required) ex: "https://eslint.org/"
 - field test_folder (optional) ex: "docker". If not set, language.lowercase() value is used
 - field config_file_name (optional) ex: ".eslintrc.yml". If not set, no default config file will be searched
-- field file_extensions (optional) ex: [".js"]. At least file_extension of file_names must be set
-- field file_names (optional) ex: ["Dockerfile"]. At least file_extension of file_names must be set
+- field file_extensions (optional) ex: [".js"]. At least file_extension or file_names must be set
+- field file_names (optional) ex: ["Dockerfile"]. At least file_extension or file_names must be set
 - method build_lint_command (optional) : Return CLI command to lint a file with the related linter
                                          Default: linter_name + (if config_file(-c + config_file)) + config_file
 - method build_version_command (optional): Returns CLI command to get the related linter version.
@@ -51,9 +51,7 @@ class Linter:
         )
         self.test_folder = None  # Override only if different from language.lowercase()
 
-        self.file_extensions = (
-            []
-        )  # Array of strings defining file extensions. Ex: ['.js','.cjs']
+        self.file_extensions = []  # Array of strings defining file extensions. Ex: ['.js','.cjs']
         self.file_names = []  # Array of file names. Ex: ['Dockerfile']
         # Default name of the configuration file to use with the linter. Ex: '.eslintrc.js'
         self.config_file_name = None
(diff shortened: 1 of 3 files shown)


nvuillam (changes requested):
Great job with this PR about performances & enhancements :)

Please can you update build.py to add the new variables in the linters documentation ?
It's around there : https://github.com/nvuillam/mega-linter/blob/master/.automation/build.py#L502

nvuillam:
+ some debug to perform, apparently it broke some existing behavior:

https://github.com/nvuillam/mega-linter/pull/139/checks?check_run_id=1523099089

For example it should not consider .dockerfilerc as a file for Perl linting

Maybe you could bring back  he pattern with all the checks calling `continue` if failed, then only at the end add the file in the list ?
https://github.com/nvuillam/mega-linter/blob/master/megalinter/Linter.py#L415

megalinter/Linter.py
```diff
@@ -414,7 +410,16 @@ def load_reporters(self):
     # Collect all files that will be analyzed by the current linter
     def collect_files(self, all_files):
         # Filter all files to keep only the ones matching with the current linter
+
+        file_extensions = set(self.file_extensions)
+        file_names_regex = f'^{"$|^".join(self.file_names)}$'
```
kolomenkin, line 415: Please not here I emulate "full match" for regular expressions. Prepend by `^` and append `$`.
So "Dockerfile" will match only "Dockerfile".

nvuillam: about Dockerfile, why not directly set these regexes in the dockerfile.yml descriptor ?

kolomenkin: 1. I understood the problem in my code (I broke checking file content and some other logic).
Now I have fixed.

2. My implementation allows specifying regular expression inside of dockerfile.yml descriptor.
I know 2 ways of applying regular expression to a string:
- search for regex matching substring
- check if the whole string matches regex. I.e. the whole text should be matched.

I chose the second way because:
- many users will forget to add `^` and `$`.
- adding `^` and `$` twice does not change logic
- no need to modify anything if you need to match by full file name like `Dockerfile`
- it still allows partial match by adding `.*` around the regex

nvuillam: I agree with your choice of option 2 :)

nvuillam:
Test cases are passing, but you have an abusive copy-paste detected :p 

https://github.com/nvuillam/mega-linter/pull/139/checks?check_run_id=1523837093

image


+ please can you update build.py to your new variables are added in the variables table of each linter documentation ?
https://github.com/nvuillam/mega-linter/blob/master/megalinter/Linter.py#L415

kolomenkin:
> * please can you update build.py to your new variables are added in the variables table of each linter documentation

1. About code duplication. I tried hard to make the code unified (duplicated) as much as possible. Previously we had almost the same code twice. But it was written in different styles.

2. I will update documentation.

nvuillam:
About copy paste, i think if you factorize that in an utils class method, it would be ok to not trigger the copy-paste blocking alert :)

image

nvuillam:
+ CHANGELOG please :)

kolomenkin:
I will do it tomorrow

kolomenkin:
I'm looking the code and found inconsistent functionality with `file_contains`. It is a list of regexes. But sometimes YAML author fills it as a list of strings.

Correct:
```
file_contains:
  - "AWSTemplateFormatVersion"
  - "(AWS|Alexa|Custom)::"
```

Mistake:
```
file_contains:
  - "schema.management.azure.com"
```

Should be:
```
file_contains:
  - "schema\\.management\\.azure.com"
```
or 
```
file_contains:
  - 'schema\.management\.azure.com'
```

1. I wonder which way is better?
2. I propose to rename `file_contains` to `file_contains_regex` to make it less confusing.

kolomenkin:
> I'm looking the code and found inconsistent functionality with `file_contains`. It is a list of regexes. But sometimes YAML author fills it as a list of strings.
> 
> Correct:
> 
> ```
> file_contains:
>   - "AWSTemplateFormatVersion"
>   - "(AWS|Alexa|Custom)::"
> ```
> 
> Mistake:
> 
> ```
> file_contains:
>   - "schema.management.azure.com"
> ```
> 
> Should be:
> 
> ```
> file_contains:
>   - "schema\\.management\\.azure.com"
> ```
> 
> or
> 
> ```
> file_contains:
>   - 'schema\.management\.azure.com'
> ```
> 
> 1. I wonder which way is better?
> 2. I propose to rename `file_contains` to `file_contains_regex` to make it less confusing.

Sorry. I found in the code strings are distinguished from regexes by first char. This is very unusual. And needs good documentation. It would be much better to allow regular expressions only.

nvuillam:
Agreed with property renaming & updating existing regexes in descriptor files :)
Plz also rename in build.py, file_contains is also used to built markdown doc

kolomenkin:
> Agreed with property renaming & updating existing regexes in descriptor files :)
> Plz also rename in build.py, file_contains is also used to built markdown doc

I have sent a email to you

PR #148, opened 2020-12-12
Allow matching filenames by regular expressions

kolomenkin:
Fixes #

## Proposed Changes

- Improve caching of compiled regular expressions

- Descriptor yaml schema changes:
  - Rename `files_names_not_ends_with` to `file_names_not_ends_with`
  - Rename `files_names` to `files_names_regex` and change behavior to expect regular expressions in the list.
    They are applied using full match (i.e. the whole text should match the regular expression)

- Configuration parameters changes:
  - Add config setting: `<LINTER_NAME>_FILE_EXTENSIONS` to override corresponding value from linter descriptor file
  - Add config setting: `<LINTER_NAME>_FILE_NAMES_REGEX` to override corresponding value from linter descriptor file

### Author/Contributor

### Reviewing Maintainer


nvuillam (changes requested):
That's pretty elegant :)
Please could you:
- Update build.py to add XXX_FILE_NAMES and XXX_FILE_EXTENSIONS in the linters docs ? ( with descriptor values in "default" column)
- Update changelog then build

PR #320, opened 2021-02-08
Add pathchk on pathnames in git

ktomk:
## Proposed Changes

1. Add pathchk on pathnames in git

### Author/Contributor

### Reviewing Maintainer


ktomk:
Do you have some documentation somewhere how to build and test? @nvuillam 

I have hot-patched some of the python scripts as they didn't ran out of the box with my local python3 despite running

~~~
python3 -m pip install --user -r requirements.dev.txt
~~~

first. In the end I could run `./build.sh` with exit status 0 and could see a lot of files are getting generated and what puzzles me is that these are all tracked in the repository - do I need to add them to the PR? Please confirm.

nvuillam:
Yes you can add the build result in a second commit, it contains auto generated doc & test classes :)
To debug, just right click on the generated test class in pycharm :)

ktomk:
I have no pycharm here so a command line would be helpful, not a regular pythonista here.

And was that a joke or for real to add these files? Sorry, I'm really puzzled about that.

ktomk:
Ah and another question. This check is specifically to git. Instead of checking files and their contents, it checks all pathnames of all files _in the repository_.

I write this explicitly because that is not the working tree and I would like to disable the generation of `GIT_GIT_PATHCHK_FILTER_REGEX_INCLUDE` and similar configuration directives which I assume do not work because of that.

The check either passes with exit status 0 or fails with exit status non-zero (123 normally).

The output gives a report about affected paths.

nvuillam:
> Another question / remark: This check is specifically to git. Instead of checking files and their contents, it checks all pathnames of all files _in the repository_.
> 
> I write this explicitly because that is not the working tree and I would like to disable the generation of `GIT_GIT_PATHCHK_FILTER_REGEX_INCLUDE` and similar configuration directives which I assume do not work because of that.
> 
> The check either passes with exit status 0 or fails with exit status non-zero (123 normally).
> 
> The output gives a report about affected paths.

This is not the first linter that ignores XXX_FILTER_REGEX_INCLUDE & XXX_FILTER_REGEX_EXCLUDE, this is the case for most of linters not linting file by file but from a folder

I think the key is to update build.py, and to generate in doc XXX_FILTER_REGEX_INCLUDE & XXX_FILTER_REGEX_EXCLUDE only when linter.cli_lint_mode not in ["file","list_of_files"] :)
-

megalinter/descriptors/git.megalinter-descriptor.yml
```diff
@@ -24,4 +24,20 @@ linters:
...
+  # GIT_PATHCHK
+  - linter_name: git_pathchk
+    lint_all_files: true
+    test_folder: git_pathchk
+    linter_text: Git files non-portable pathname check
+    linter_url: https://git-scm.com
+    linter_repo: https://github.com/git/git
+    linter_megalinter_ref_url: never
+    cli_config_arg_name: ""
+    cli_executable: /bin/sh
```
nvuillam, line 36: I'd prefer executable to be git , and other args managed in cli_lint_extra_args, so that debugging in windows would work too, please could you manage that ? :)

nvuillam (changes requested):
Plz also update CHANGELOG :)

megalinter/descriptors/git.megalinter-descriptor.yml
ktomk: Yes, also my preference, but in this case not possible. a) It is  complex shell command, hence the shell to execute it. b) I have no Windows test-system and I assume despite Windows has some kind of Posix certification, the Posix Pathchk utility is not available on that platform. c) The Mega-Linter container is Linux.

ktomk: @nvuillam Btw, how can I execute the configured linker on bare-metal? Or do I need to build the 6-7 GB container first? This linter does not need any new binary, I can mount the repo into the container if it helps.

ktomk: > Plz also update CHANGELOG :)

I have no clue how that works. Do you have any documentation at hand?

Also do you have any developer / contributing documentation? Install from source? Develop the source version? Anything?

nvuillam: Contributing guide is here :)
https://nvuillam.github.io/mega-linter/contributing/

If infos are missing, just ask ^^

ktomk: Thanks, I was missing this doc, that helps a lot.

ktomk:
Still having this on my list, it needs some project setup first on my end, then will update here. @nvuillam

nvuillam:
It does not want to... i'm afraid you have to create a new PR ^^

ktomk:
I restored the original head, let me know if it helps.

nvuillam:
reopened, but with conflicts ^^

ktomk:
conflicts? which conflicts? ^^

the intermediate problem from yesterday seems gone. I just rebuild and then the megalinter-configuration.jsonschema.json went back to the old indentation.

converted to draft as this is not for a merge right now.

ktomk:
This is still in preparation it just takes a bit longer as while pathchk is indeed nice, it may be a bit of a hindrance for some projects and need a fallback (working on it needs some more tests and likely some extra tooling).

For starters currently considering the more basic ASCII  character check what can be [seen in the pre-commit hook sample](https://github.com/git/git/blob/master/templates/hooks--pre-commit.sample#L24-L46).

A true pathchk can then add on top of it.

nvuillam:
Please can you explain what this linter checks ? :)
I'm sorry i'm not aware of pathcheck :/

ktomk:
That filenames (and more specifically path segments) are valid and portable. You can find a description with [`pathchk(1)`](http://www.gnu.org/software/coreutils/manual/html_node/pathchk-invocation.html). Makes sense @nvuillam?

nvuillam:
Crystal clear, thanks for the explanation :)

nvuillam:
@ktomk is your PR still in draft ? ^^

ktomk:
yes it is .... and I still need to do some clean-ups and add an option that it is lax by default as otherwise the maximum path segment length is 14 and this will bring in rage as for today's kids this I'm pretty sure is not well understood.

I have control via .gitattributes which I think is a good configuration file for a git repository, but good defaults especially in non-interactive use like in mega-linter normally pay off.

also I'm wondering if applying on the whole repository is always beneficial on mega-linter and not just on the paths of the change-set (added and edited paths). I can't find the docs for this, I remember there are some, like how to optimize the invocation and/or some environment parameters. maybe you have it at hand?

nvuillam:
@ktomk > MegaLinter is able to work on the list of updated files or the whole list of files, and i'm not enough git expert to know the detailed command for max 14

I'm pretty sure MegaLinter itself is full of path of more than 14 chars, so we can describe pathchk as "format" so in case it fails, it will be just a warning but will not make the whole MegaLinter fail ^^

ktomk:
bump ti di li bump bumply

PR #381, opened 2021-03-30
Use Python virtual-environment in dev-dependencies shell example

ktomk:
Just a small one to get hands warm.

PR #385, opened 2021-03-31
Fixup #381: Wrong file edited

ktomk:
Changes in a4996d4 are overwritten by the build.

ref: a4996d48c6ce5932b4decb71b38712a981dc4cef

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -14,8 +14,25 @@ We're thrilled that you'd like to contribute to this project. Your help is essen
 - Run the following commands at the root of the repository to install required dev dependencies
 
 ```shell
+python3 -m venv venv
+source venv/bin/activate
+echo "venv/" >> .git/info/exclude
 python3 -m pip install -U pip
 python3 -m pip install -r requirements.dev.txt
+```
+Second level dev dependencies are installed by running `./build.sh` which is also a test if the installation worked
+```shell
+./build.sh
+2021-03-30 19:40:03,790 [INFO] Validating ansible.megalinter-descriptor.yml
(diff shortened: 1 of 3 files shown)


nvuillam (approved):

nvuillam:
hmmm hole in my memory, this issue has been solved in another PR, right ?

ktomk:
> hmmm hole in my memory, this issue has been solved in another PR, right ?

Yes, I kept the original PR number here as reference.

We did not solve it cause the wrong file edited (the generated one only).

just refreshed against master, but need to do again if I can trust this interface here while typing.

Let me take care.

ktomk:
Refreshed against master.

nvuillam:
Merged without the 40mn CI :D (bad practice but it's only MD so no risk)

At worse it will force me to update .cspell.json ^^

ktomk:
yeah that .cspell.json ^^ 

I can also take a look.  let's see for whats next ... ^^

nvuillam:
If you really have time.... https://github.com/nvuillam/mega-linter/issues/376

You could integrate your first linter in MegaLinter : https://github.com/todogroup/repolinter ;)

ktomk:
@nvuillam what you say ^^  I wanted to make first of all my first linter and then I was thinking about to add PHP-CS-Fixer or others in the PHP section.

nvuillam:
If you are a PHP guy, go for a PHP linter ;)
Just make sure that if there are several formatters/fixers , they don't fight between each other about the format ^^ (and if they do, you have to provide a variable to let the user choose a formatter, like with MARKDOWN_DEFAULT_STYLE :) )

PR #386, opened 2021-03-31
Keep license pre-formatted in docs

ktomk:
the license page didn't look good due to intermixing markdown with pre-formatted plain text of the LICENSE file.

loading it in by snippet instead of copying verbatim solves this.

LICENSE can stay pre-formatted within a code-block in the Markdown.

adds the pymdownx.snippets mkdocs extension.

Currently online: https://nvuillam.github.io/mega-linter/license/

Selection_138


nvuillam (approved):
Not tested but I'm pretty sure it will work ;) ( we'll see that in 2 mn ^^ )

ktomk:
> Not tested but I'm pretty sure it will work ;) ( we'll see that in 2 mn ^^ )

Looks beautiful now. https://nvuillam.github.io/mega-linter/license/

Btw. could we say that symbolic links are a given on all build systems? If so mkdocs would support that as well instead of the copy operations.

nvuillam:
@ktomk during the copy there may be some dynamic stuff, so not sure it's safe to change that :p 
But why not ^^

nvuillam:
it's really much better, nice job ! :)

image

PR #390, opened 2021-03-31
Preserve formatting (minor)

ktomk:
via:

    $ wget https://www.gnu.org/licenses/agpl-3.0.txt -O - >| LICENSE

now after preformatting works, you see whats not so beautiful with the actual document ...


nvuillam:
Even better like that yep, thanks :)

PR #418, opened 2021-04-18
Check _RULES_PATH for active_only_if_file_found check 

### LINKED ISSUE #374: Difficulty using linters that define 'active_only_if_file_found' inside a monorepo
I am seeing an issue in the case of JSX in a monorepo but maybe there's an obvious solution. The structure is `.mega-linter.yml` at the root level with multiple subdirectories inside. The ` .eslintrc.json` is defined inside one of the subdirectories.

```
.
├── .mega-linter.yml
└── backend
└── webapp
    └── .eslintrc.json
```

This is the command I was attempting:
```
	docker run -t \
		-v ${PWD}:/tmp/lint \
		-e ENABLE_LINTERS=JSX_ESLINT \
		-e FILTER_REGEX_INCLUDE=webapp \
		nvuillam/mega-linter-javascript:v4
```
(using FILTER_REGEX_INCLUDE seems to work fine with the backend cases)

The issue I am seeing is that some linters, like the JSX one, specify [active_only_if_file_found](https://github.com/nvuillam/mega-linter/blob/master/megalinter/descriptors/jsx.megalinter-descriptor.yml#L18) and the [file check](https://github.com/nvuillam/mega-linter/blob/c3312dc35e70c565650475476a6fe0fe5c8a01f8/megalinter/Linter.py#L238) does not take into account other hints that could point to this file i.e. `JSX_ESLINT_CONFIG_FILE` or a `LINTER_RULES_PATH` containing a `.eslintrc.json` file. We could move `.eslintrc.json` to the root but that would take that file out of the project folder that contains `package.json`. We could mount the `webapp` subdirectory as the root workspace but this would omit the top-level `.mega-linter.yml` file from being utilized.

So what is left is manipulating the view of the directory structure through something like this:

```
	docker run -t \
		-v ${PWD}/webapp:/tmp/lint \
		-e ENABLE_LINTERS=JSX_ESLINT \
		-v ${PWD}/.mega-linter.yml:/tmp/.mega-linter.yml \
		-e MEGALINTER_CONFIG=../.mega-linter.yml \
		nvuillam/mega-linter-javascript:v4
```
What we are doing:
- we mount the "webapp" folder into the default workspace /tmp/lint. This finds the `.eslintrc.json` file but we no longer have the top-level .mega-linter.yml file.
- we mount `.mega-linter.yml` file into the /tmp directory (one level above the workspace).
- we set MEGALINTER_CONFIG to look one-level above from the workspace to find it.

This works but feels hacky. As far as I can tell, this only affects linters that have `active_only_if_file_found` defined.

Or maybe there is something I'm missing?

omusavi:
Currently, active_only_if_file_found only checks the root workspace folder for the required files. These files may not always be present at the root, especially in a monorepo scenario. By looking in the `_RULES_PATH` folder, we can enable these linters to run as expected.

Fixes #374 

## Proposed Changes

1. Respect LINTER_RULES_PATH when checking for required files to turn on certain linters (ESLint)
2. When setting `config_file` and `linter_rules_path` is set, make it relative to `self.workspace` (might only be required for local execution where CWD is `/`)

### Author/Contributor

### Reviewing Maintainer


omusavi:
First contribution, saw an issue that I am also running into so wanted to try and submit a fix! My concern is mostly around the 2nd commit where i am appending the `linter_rules_path` to `workspace`. It seems like a safe change to me (I am guessing linter rules are always relative to the workspace root?). I think this also only affects local runs since CWD is set differently when run with `mega-linter-runner` vs in github actions. 

Marked as draft to make sure this is ok before merging

omusavi:
Also looks like the linter is failing in a file I did not modify, should I fix as a part of this PR or is it not something to worry about?

nvuillam:
Sorry gor the delay, i'll have a look asap :)

nvuillam:
Please can you try to merge master in your branch ?
+ add your update in root changelog.md
+ bash build.sh

Thanks :)

omusavi:
@nvuillam Updated! `build.sh` did not generate any new files to update. Assuming you feel this change is safe, I am setting it to ready for review

nvuillam:
@omusavi CI is not happy... i think I know why :)
Default rules can be found in .github/linters, and you replaced the statement, whereas you should add another one :)

image

omusavi:
Sorry I am not sure I understand what you mean. I don't think I replaced any checks, in the code in the image above I just prepended the workspace path to the linter rules path which is a workaround to get it working locally. 

I guess I am a bit confused about the CWD in these cases (also called out in #419), are we expecting `.github/linters` to be in the workspace root?

nvuillam:
@omusavi yes github/linters must be in workspace root :)

Basically, let the elif in ref like it was before, and add an elif before with the code in green, and that should do the trick :)

image

omusavi:
Ok I still do not completed understand under what scenarios the file at `self.linter_rules_path + os.path.sep + self.config_file_name` exists but `self.workspace + os.path.sep + self.linter_rules_path + os.path.sep + self.config_file_name` doesn't if linter_fules_path is always inside self.workspace...perhaps absolute vs relative path then? 

Added the change as you directed and it looks like CI is happy, so this is just for my own knowledge!

megalinter/Linter.py
```diff
@@ -326,7 +329,8 @@ def load_config_vars(self):
         # 1: http rules path: fetch remove file and copy it locally (then delete it after linting)
         # 2: repo + config_file_name
         # 3: linter_rules_path + config_file_name
-        # 4: mega-linter default rules path + config_file_name
+        # 4: workspace root + linter_rules_path + config_file_name
```
omusavi, line 332: I just want to confirm this is the correct interpretation here...

nvuillam:
> Ok I still do not completed understand under what scenarios the file at `self.linter_rules_path + os.path.sep + self.config_file_name` exists but `self.workspace + os.path.sep + self.linter_rules_path + os.path.sep + self.config_file_name` doesn't if linter_fules_path is always inside self.workspace...perhaps absolute vs relative path then?
> 
> Added the change as you directed and it looks like CI is happy, so this is just for my own knowledge!

Probably in MegaLinter own test classes :)

nvuillam (approved):
Great, thanks for your contribution :)

PR #450, opened 2021-05-11
Make FILE_NAMES_REGEX match against the full filename (with extension)

### LINKED ISSUE #445: `*_FILE_NAMES_REGEX` only examines the file name and doesn't include the extension
**Describe the bug**
When setting e.g. `BASH_SHELLCHECK_FILE_NAMES_REGEX`, it only looks at the file name and not the extension, even if an extension is present. I would like to be able to tell Mega-Linter to check the file `foo` as a Bash file, but not `foo.py`. This doesn't appear to be possible with the current implementation.

**To Reproduce**
I've created a branch that demonstrates this behavior. See https://github.com/ottumm/mega-linter-test/pull/2

You can see ShellCheck attempting to check a `.py` file here: https://github.com/ottumm/mega-linter-test/pull/2/checks?check_run_id=2530002709#step:4:133

**Expected behavior**
`*_FILE_NAMES_REGEX` should use the entire filename, including extension, when evaluating, so that I can use e.g. `BASH_SHELLCHECK_FILE_NAMES_REGEX: ["^foo$"]` to match the file `foo` and not the file `foo.py`.

ottumm:
Fixes #445

## Proposed Changes

This changes the behavior of `*_FILE_NAMES_REGEX` settings to match against the full filename (including extension).

### Author/Contributor
- [x] If documentation is needed for this change, has that been included in this pull request. _No documentation changes required._

### Reviewing Maintainer


ottumm:
Gotcha, makes sense and sounds good! I'll update the PR.

ottumm:
> I have an issue with CI (updated linters)

Those pesky linters, huh? ;-)

nvuillam (approved):

ottumm:
> Thanks a lot for your contribution :)

My pleasure — thank you for your work maintaining this!

PR #452, opened 2021-05-12
Add terraform fmt

### LINKED ISSUE #447: New Linter suggestion: `terraform fmt`
The current linter suite includes a call to `terragrunt hclfmt`, but this only formats Terragrunt HCL syntax files, not standard Terraform files.

A very similar linter that calls `terraform fmt` on relevant files seems appropriate.

None

**Additional context**
I'd be happy to contribute this, but I wanted to first check with the maintainers (@nvuillam).

I'm also unsure of the name. The existing pattern of `TERRAFORM_TERRAGRUNT` would lead me to `TERRAFORM_TERRAFORM`, but they both should really be something more like `TERRAFORM_TERRAGRUNT_HCLFMT` and `TERRAFORM_TERRAFORM_FMT`.

tpansino:
Resolves #447

## Proposed Changes

1. Adds a new `terraform fmt` formatter for Terraform files

### Author/Contributor

### Reviewing Maintainer


tpansino:
@nvuillam, 2 questions:

1. If the linter command is `terraform fmt <file>`, should the `linter_name` in the descriptor definition be `terraform-fmt`, or `terraform_fmt`? It's currently the first one, but I see you have a `git_diff`, and I'm wondering if it's incorrect.
2. Why do contributors run `build.sh` and commit the results, instead of that step being part of the release automation? It seems to generate a lot of noise in the PR diff.

tpansino:
One more question 😃 - how do I get the correct version of Terraform to show up in the docs? I set `version_extract_regex` and re-ran the `build.sh`, but it didn't update. The regex is pretty simple, so I believe it's correct.

nvuillam:
@tpansino 

> 1. If the linter command is `terraform fmt <file>`, should the `linter_name` in the descriptor definition be `terraform-fmt`, or `terraform_fmt`? It's currently the first one, but I see you have a `git_diff`, and I'm wondering if it's incorrect.

That's ok, - are replaced by _ to build configuration variables anyway :)

> 2. Why do contributors run `build.sh` and commit the results, instead of that step being part of the release automation? It seems to generate a lot of noise in the PR diff.

Because build.sh also generates test classes, and I can not consider a PR ready if the documentation is not correctly generated :)
It indeeds makes a lot of updated files when there is a new linter... but that's the cost to have 100% auto-generated doc, and that's a cost that I'm willing to pay :D
For tiny PRs like quick fixes I sometimes accept that build.sh is not run, and in that case I have another job that will take care of that later ^^

> 3. One more question 😃 - how do I get the correct version of Terraform to show up in the docs? I set version_extract_regex and re-ran the build.sh, but it didn't update. The regex is pretty simple, so I believe it's correct.

It is extracted from https://github.com/nvuillam/mega-linter/blob/master/.automation/generated/linter-versions.json , which is generated by another job, don't worry, if the test case passes, it will be ok and Mega-Linter will auto-update its doc with the good version :)

nvuillam:
About 1 & 3 answers, I'm referring to auto-update job -> https://github.com/nvuillam/mega-linter/blob/master/.github/workflows/auto-update-linters.yml

It runs daily, and it:
- Builds a new Mega-Linter docker image, with linters latest versions
- Collect linters version & help from the newly build image
- If diff in versions, create a PR
  - If PR job is successful, PR is auto-merged in master
     - Doc rebuild & deployment is performed by a job that detects a push in master branch 
     - New @latest / @insiders version of Mega-Linter is generated by a job that detects a push in master branch

That's such architecture that allows me to just look at Mega-Linter maintaining itself most of the times, except when I release a new official (V4) version , or like in this case when there are nice evols proposed by nice contributors :)

tpansino:
@nvuillam 2 more questions :)

1. I can't seem to get the link check to pass in CI. Is it supposed to fail for new linter additions?
2. Is there any documentation on the testing? From the existing files, I assume one `good` file is required which is correctly linted/formatted, and one `bad` which is not? When would multiple files be useful?

nvuillam:
Sometimes a build fails when one of the linters install commands fails
In your case `06:54:12| wget: bad address 'alpine-pkgs.sgerrand.com' `
I just tested the address, it should work now, you can try again :)

About test cases, one good and one bad file are enough, sometimes you can have several if you consider that different good or bad cases should be tested to be sure the linter works :)

nvuillam:
About doc, there is one line about the 2 test files, if you think there should be more please feel free to enhance the doc in your PR  :)

https://nvuillam.github.io/mega-linter/contributing/#add-a-new-linter

tpansino:
@nvuillam Is there anything else needed for this one, or can it be merged? Trying to reduce the number of PRs I have open :)

nvuillam:
@tpansino it will be merged when the CI will pass ;) 
For the last attempt, there has been again a linter install url ot responding, you need to trigger the job again

tpansino:
Thanks for the merge @nvuillam. 🙏

I assume you noticed the final check wasn't passing? That was what I was referring to in 1 [here](https://github.com/nvuillam/mega-linter/pull/452#issuecomment-841135887) :) It seems because the docs are committed, for new linters, the links pointing to `master` aren't valid yet, so they fail the link check.

I completely agree with your points about automation. In future iterations of the automation though, it may be beneficial to split out the doc generation from the test case generation 🤷 that way checks are always passing before merge, and auto-generated docs are not creating extra PR noise in the review. Something to think about 👍

nvuillam:
@tpansino yep, it happens than for some PR's, there are false positives :) ( but never on `Build & Deploy - DEV`, which is the only job that must never be bypassed ^^

I fully agree with you, there is clearly no point noising so many files... and maybe we could go further: why committing automated updates in /docs when they can be directly generated then pushed in gh_pages branch ? :)

PR #458, opened 2021-05-12
Exclude .terraform, .mypy_cache, and REPORT_OUTPUT_FOLDER

tpansino:
## Proposed Changes

1. Excludes `.terraform` and `.mypy_cache` folders by default, as these are cache folders for Terraform and Python's `mypy` tool respectively
2. Exclude the `REPORT_OUTPUT_FOLDER` dir based on the config, instead of using a hard-coded `report` value

### Author/Contributor

### Reviewing Maintainer


nvuillam:

PR #463, opened 2021-05-13
Fix anchor link typo

kidonng:
## Proposed Changes

It seems GitHub uses dashes instead of underscore for the anchor link:

image

### Author/Contributor

### Reviewing Maintainer


nvuillam (approved):
you're right, thanks :)

PR #470, opened 2021-05-15
Update `rstfmt` lint/fix args

tpansino:
## Proposed Changes

`rstfmt` was updated about [8 months ago](https://github.com/dzhu/rstfmt/commit/6214976566e304cd8c89e376ff1cbb035c90a340) with a breaking change to the CLI switches. This version also made it into Mega-Linter (though I'm not exactly sure how it got past the tests. It's completely broken from what I can tell...)

This PR:
- Updates `rstfmt` definition to use `--check` for linting and just `rstfmt` for fixing.
- Adds version extraction to `rstfmt`
- Enables `rstfmt` (not sure why it was disabled?)

### Author/Contributor

### Reviewing Maintainer


tpansino:
Looks like the `rstfmt` tests didn't have a "bad" test case, I'm guessing that's why they never caught that the tool was broken. 🤷

nvuillam:
@tpansino i disabled this formatter to not block other PRs , your debug is much cleaner solution ;)
But it seems the test success still fails :/
To avoid long CI validation, you can debug locally by installing rstfmt and run test class dirrctly from pycharm or vscode with latest Microsoft python extension :)

tpansino:
Looks like `rstfmt` is broken due to Sphinx 4.0 release. Filed a bug ticket https://github.com/dzhu/rstfmt/issues/12. This PR won't pass until that gets fixed unfortunately :(

nvuillam:
Did it work ? :)

tpansino:
Not exactly, as you can see :)

Looks like the issue is that the `<` character needs to be escaped or quoted in the Dockerfile. Let's see if it builds and tests pass now...

tpansino:
@nvuillam, it looks like this worked!

Remind me, is there a way to download the test Docker image and manually verify it?

nvuillam:
If CI passed it means that it should be ok :)
Once I validate, it will be available in @latest version , then in official V4 versions in a few days/weeks, it will let you thme time to test :)

nvuillam:
thanks for your contribution :)

PR #487, opened 2021-05-25
Allow applying fixes on push events

### LINKED ISSUE #481: Applying fixes does not work
**Describe the bug**

The only way of applying fixes that seems to work is the commit mode on pull request events from one branch to another in the same repository. All other cases fail for various reasons.

**To Reproduce**
* Create a new branch.
* Commit changes that require fixes.
* Push to the remote branch.
* Create a pull request to the main branch.
* See the log.
* Repeat for all combinations of `APPLY_FIXES_EVENT` and `APPLY_FIXES_MODE`.

**Expected behaviour**
Fixes are applied as commit or pull request (according to `APPLY_FIXES_MODE`) for push or pull request events (according to `APPLY_FIXES_EVENT`).

**Additional context**
* Applying of fixes on push events is blocked by the condition `github.event.pull_request.head.repo.full_name == github.repository` which can never be satisfied because `github.event.pull_request` exists only for pull request event.
* Applying of fixes in the pull request mode does not work because the repository is checked out in a detached head and the `peter-evans/create-pull-request` action complains with:
```
fatal: ref HEAD is not a symbolic ref
Error: When the repository is checked out on a commit instead of a branch, the 'base' input must be supplied.
```
I tried to run the workflow with the branch `HEAD` checked out using: `ref: ${{ github.event.pull_request.head.sha }}` in the `checkout` action but that makes the Mega-Linter runner fail with:
```
Warning: Unable to pull current branch <SHA>
cmdline: git diff --name-only master...<SHA>
stderr: 'fatal: Invalid symmetric difference expression master...<SHA>
```
where `<SHA>` is a commit SHA which is different from the branch `HEAD` at which the repository was initially checked out.
(Could it be the commit with the fixes?)

**How about pull requests between repositories?**
Besides fixing these issues it would also be very useful if the fixes could be applied in pull request events from a fork to an upstream repository, since that is the way pull requests are mostly used.

**I don't know if it is related, but for completeness**
For tests of pull requests from a fork to an upstream repository, this warning always appears:
```
Listing updated files in [/github/workspace] using git diff, then filter with:
Warning: Unable to pull current branch <SHA>
Git diff :
```
where `<SHA>` corresponds to the merge commit created by the checkout action.
This is then followed by this error for each linter:
```
[GitHub Status Reporter] Error posting Status for <CATEGORY>with <linter>: 403
GitHub API response: {"message":"Resource not accessible by integration","documentation_url":"https://docs.github.com/rest/reference/repos#create-a-commit-status"}
```

vkucera:
Partially fixes issue #481 

## Proposed Changes

1. Modify the condition for executing the following workflow steps:
* Create Pull Request with applied fixes
* Create PR output
* Prepare commit
* Commit and push applied linter fixes

so that it can be satisfied for push events.

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -128,6 +128,7 @@
         "Klingenberg",
         "Ktlint",
         "Kubernetes",
+        "Ku\u010dera",
         "LASTEXITCODE",
         "LINTR",
         "LOCALFOLDER",
--- a/.cspell.json
+++ b/.cspell.json
@@ -128,7 +128,7 @@
         "Klingenberg",
         "Ktlint",
         "Kubernetes",
-        "Ku\u010dera",
+        "Kučera",
         "LASTEXITCODE",
         "LINTR",
         "LOCALFOLDER",
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -23,7 +23,7 @@ python3 -m pip install -r requirements.dev.txt
 _If it does not work, just run the following script_
 
 ```shell
-pip install -r requirements.dev.
+pip install -r requirements.dev.txt
 pip install mkdocs-material
 npm install markdown-table-formatter -g
 ```
@@ -89,9 +89,9 @@ Then run `bash build.py` and it will generate all the rest !
 - Test classes
 - Configuration JSON schema
 - Online documentation menus
-  
+
 ![Screenshot](https://github.com/nvuillam/mega-linter/blob/master/docs/assets/images/ContributingAddLinter_1.jpg?raw=true>)
-  
+
 
 ### CI/CT/CD
 
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -9,7 +9,7 @@ We're thrilled that you'd like to contribute to this project. Your help is essen
 
 ### Pre-requisites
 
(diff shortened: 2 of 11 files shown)


nvuillam (approved):

nvuillam:
@vkucera there is an update to add to cspell.json before I can merge :) (you can just take generated cspell.json file in Mega-Linter artifacts then commit & push )
If you are in a good mood, you can also complete the doc with the missing points you noticed :) ( else i'll do that later)

vkucera:
Hi @nvuillam , I modified also other occurrences of the same condition in files that I had overlooked initially. Can you please check that the changes are appropriate?
I added the updated `.cspell.json` but the error did not disappear. I guess it might be due to the diacritics in my name so I replaced the Unicode code of the character with the character itself. Let's see if the test passes now.
I also fixed the missing file extension in the contribution instructions.
Regarding the node.js version, I don't know which is the minimum required one so I couldn't add any meaningful instructions there. But if you tell me the version, I can add it in this PR.

nvuillam:
I'm quite sure node14 will be ok :)

vkucera:
> I'm quite sure node14 will be ok :)

Done! :-)

nvuillam:
@vkucera all good, thanks a lot for your contribution and your updates, and welcome to the team :)

vkucera:
Thanks!

PR #506, opened 2021-06-09
bump black to 21.5b2

gnawhleinad:
https://pypi.org/project/black/ 21.5b2 was released on May 31, 2021.

The current version 20.8 was released ~August 2020. `py39` was added in
September 2020 with https://github.com/psf/black/pull/1717.


gnawhleinad:
> Mega-Linter updates itself automatically , so when command `pip install black` will install this newest version, it will be taken in account :)

@nvuillam, thanks for that context!

However, I'm a bit confused since https://pypi.org/project/black/'s `21.5b2` was released on May 31, 2021. How frequently does `mega-linter` update itself?

nvuillam:
The auto update job is run daily :)
Black probably released a version but did not update the pip package yet

gnawhleinad:
> The auto update job is run daily :)
> Black probably released a version but did not update the pip package yet

Please let me know if I need to file a bug... but it seems like `black` _was_ released:

```
$ pip download black
Collecting black
  File was already downloaded /hodor/black-21.5b2-py3-none-any.whl
...
```

This seems consistent with https://pypi.org/project/black/#history which shows:

Screen Shot 2021-06-10 at 10 48 20 AM

Is it possible that `mega-linter` is looking for non-pre-release versions...?

nvuillam:
I think pip3 install retrieves the latest official release, as you can see there -> https://github.com/nvuillam/mega-linter/blob/63cd757a63af403764daefd5d5a97479225be41d/Dockerfile#L126

Pre-releases must be considered not stable enough, so it seems normal they are not used in Mega-Linter ^^

If you want to force a black version, you can use [PRE_COMMANDS](https://nvuillam.github.io/mega-linter/configuration/#pre-commands) configuration to install it your way, it is called before the linters execution :)

PR #520, opened 2021-06-23
Fix terraform fmt & terragrunt hclfmt

tpansino:
## Proposed Changes

- Fixes the auto fix functionality of both `terraform fmt` and `terragrunt hclfmt`
- Adds tests to validate both
- (Bonus) Fix version number extraction for `terraform-fmt` and `rstfmt` (I think?)

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/.automation/test/sample_project_fixes/terraform_for_fixes_1.tf
+++ b/.automation/test/sample_project_fixes/terraform_for_fixes_1.tf
@@ -1,0 +1,3 @@
+locals {
+  foo =     "bar"
+}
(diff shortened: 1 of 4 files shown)


tpansino:
@nvuillam , when you have a moment, can you confirm my version extraction fix is correct? I thought I had the regex set up correctly, but then looking at the generated docs it included extra text.

I don't see how to generate the `linter-versions.json` file myself, so I can't confirm if my fix worked. :sweat_smile:

nvuillam:
Will do that today :)
About linter version & help, it generates itself with ci job "Auto update linters" , which is triggered daily + at each push in master :)

nvuillam:
And to test locally, you can:
- install the related linter(s)
- run test class from vscode or pycharm

nvuillam (approved):
Many thanks for your contributions ;)

CI is passing, but fix is not tested by CI and version format neither, so please check with @latest / @insiders version if it's ok and submit another PR if not :)

PR #529, opened 2021-06-28
Fix auto-fixes for Terraform fmt, Terragrunt hclfmt, and rstfmt

tpansino:
## Proposed Changes

- Added `cli_lint_fix_arg_name: "--megalinter-fix-flag"` for Terraform fmt, Terragrunt hclfmt, and rstfmt formatters
- (Bonus): Added `cli_config_arg_name` for Checkov

### Author/Contributor

### Reviewing Maintainer


tpansino:
@nvuillam, it took me a lot of effort to figure out that these three formatters needed [this](https://github.com/nvuillam/mega-linter/blob/3b82e916cb2dcd43a7ab14a746f5a15156a5e28e/megalinter/Linter.py#L781) workaround.

It seems like this should be simpler, or at least better documented. The formatters already have `is_formatter: true` set on them, couldn't that be used in [this](https://github.com/nvuillam/mega-linter/blob/3b82e916cb2dcd43a7ab14a746f5a15156a5e28e/megalinter/Linter.py#L758) conditional check, instead of looking for `cli_lint_fix_arg_name`?

I also find that debugging these sorts of issues is challenging with the current development workflow. Maybe I'm just not using your tools or flow :) but the only way I've found to get fast feedback while debugging is to `docker run` into the container shell, hack the files until the code works, and then copy the changes out by hand. It would be a lot easier to volume mount the whole git repo and be able to tweak and re-run it. :shrug:

tpansino:
Oh, one more thing - what do you think about adding some tests for the auto-fixers? I could contribute that if you suggest to me what should be added/changed. :)

nvuillam:
It's night here, but i agree with all your messages , i'll answer in details tomorrow :)

nvuillam:
@tpansino 

Debugging: Usually I just install locally the linters and run test classes, and the behaviour is quite similar to the one within the docker image :)

Tests for auto-fixer:
That's indeed a lack for Mega-Linter, the is one test class  mega_linter_2_fixes_test.py that is not up to date, and it would be much more efficient to have for example some python_to_fix.py and python_fixed.py files , and have a generic test class checking if there is a cli_lint_fix_arg_name and checking if the result is corresponding to the expected file !

Root class for tests is LinterTestRoot so such method could be added there

nvuillam:
Meanwhile, i merged this PR, such evols would be in another :)

thanks again for your contribution :)

nvuillam (commented):
good !

PR #605, opened 2021-07-30
fix list_git_ignored_files (when `IGNORE_GITIGNORED_FILES: true`)

davidB:
Fixes #608

## Proposed Changes

Fix the bug when using `IGNORED_GITIGNORED_FILES: true` (error on call to `git status --ignored` and no files ignored), by fixing the call to external process and by using `git ls-files --exclude-standard --ignored --others` instead of `git status` to have an easier output to parse to generate the list of files to ignored.

### Details

- call `repo.git.execute("git status --ignored"]) failed with

    ```sh
    Unable to list git ignored files (Cmd('g') not found due to: FileNotFoundError('[Errno 2] No such file or directory: 'git status --ignored'')
      cmdline: g i t   s t a t u s   - - i g n o r e d)
    ```

- `repo.git.execute` return a single string not a list
- output of `git status --ignored` was not parsed correctly to create a list of file

```sh
❯ git status --ignored
On branch bugs/git_list_ignored
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        modified:   megalinter/MegaLinter.py

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   megalinter/MegaLinter.py

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        shell.nix

Ignored files:
  (use "git add -f <file>..." to include in what will be committed)
        .automation/test/mega-linter-plugin-test/report/
        .automation/test/pre-post-test/report/
        .automation/test/sample_project/report/
        .automation/test/sample_project_fixes/report/
        .pytest_cache/
        .venv/
        megalinter/.pytest_cache/
        megalinter/__pycache__/
        megalinter/linters/__pycache__/
        megalinter/reporters/__pycache__/
        megalinter/tests/__pycache__/
        megalinter/tests/test_megalinter/__pycache__/
        megalinter/tests/test_megalinter/helpers/__pycache__/
        megalinter/tests/test_megalinter/linters/__pycache__/
        report/
        site/

```

### Author/Contributor
- [ ] If documentation is needed for this change, has that been included in this pull request

### Reviewing Maintainer
- [ ] Label as `breaking` if this is a large fundamental change
- [ ] Label as either `automation`, `bug`, `documentation`, `enhancement`, `infrastructure`, or `performance`

PR #658, opened 2021-08-17
Fix git ignore functionality

### LINKED ISSUE #608: Unable to list git ignored files when `IGNORED_GITIGNORED_FILES: true`
**Describe the bug**

When using `IGNORED_GITIGNORED_FILES: true` (error on call to git status --ignored and no files ignored), by fixing the call to external process and by using git ls-files --exclude-standard --ignored --others instead of git status to have an easier output to parse to generate the list of files to ignored.

```
    Unable to list git ignored files (Cmd('g') not found due to: FileNotFoundError('[Errno 2] No such file or directory: 'git status --ignored'')
      cmdline: g i t   s t a t u s   - - i g n o r e d)
```

**To Reproduce**

1. Go to a local project (eg mega-linter) with `.gitignore` files and files ignored by .gitignore (eg: `venv`, `node_modules`, `target`,...)
2. Add a `.mega-linter.yml`

    ```sh
    echo "IGNORE_GITIGNORED_FILES: true" > .mega-linter.yml
    ```

3. run mega-linter on local project

    ```sh
    docker run --rm -it -v /var/run/docker.sock:/var/run/docker.sock:rw -v $PWD:/tmp/lint:rw nvuillam/mega-linter:v4
    ```

  - error is reported
  - files are not ignored (so lot error could be reported by third party,...)

In this sample, run in on a rust project

**Expected behavior**

Ignoring files following the `.gitignores` rules

**Screenshots**

```sh
> docker run --rm -it -v /var/run/docker.sock:/var/run/docker.sock:rw -v $PWD:/tmp/lint:rw nvuillam/mega-linter-rust:v4
...
Listing all files in directory [/tmp/lint], then filter with:
- File extensions: , .R, .RMD, .Rmd, .bash, .c, .c++, .cc, .cdxml, .clj, .cljc, .cljs, .coffee, .cpp, .cs, .css, .cu, .cuh, .cxx, .dart, .dash, .edn, .env, .feature, .go, .gradle, .graphql, .groovy, .gvy, .h, .h++, .hcl, .hh, .hpp, .htm, .html, .hxx, .java, .js, .json, .json5, .jsonc, .ksh, .kt, .kts, .lua, .md, .nf, .p6, .php, .pl, .pl6, .pm, .pm6, .pp, .proto, .ps1, .ps1xml, .psd1, .psm1, .psrc, .pssc, .py, .r, .raku, .rakumod, .rakutest, .rb, .rs, .rst, .saas, .scala, .scss, .sh, .smk, .sql, .t, .tex, .tf, .ts, .vb, .xml, .yaml, .yml
- File names (regex): Dockerfile, Jenkinsfile, Snakefile
Unable to list git ignored files (Cmd('g') not found due to: FileNotFoundError('[Errno 2] No such file or directory: 'git status --ignored'')
  cmdline: g i t   s t a t u s   - - i g n o r e d)
Kept [1492] files on [2806] found files
...
```

**Additional context**

TimPansino:
Fixes #608

## Proposed Changes

1. Fix git ignore command typing (cmdline arguments should be lists of strings, not a string with spaces)
2. Change git ignore command to use `git ls-files -io --exclude-standard` which lists all individual untracked files ignored by git. Removes the need for globbing which fails when a folder isn't listed with a trailing slash.
3. Properly split and return list of files rather than a giant string with no effect.

### Author/Contributor

### Reviewing Maintainer


TimPansino:
Tested locally and working as intended on an example with git ignored files and folders.

nvuillam:
@tpansino seems great, thanks !
But please can you replace the list of ignored files by the number of ignored files ? it seems heavy in the logs, and can even make crash some stuff as u can see in CI

image

I also see that @davidB has a different approach in PR https://github.com/nvuillam/mega-linter/pull/605 , maybe you could have a look and tell what you think about it ?

Thanks a lot ! :)

davidB:
#605 
- also uses `git ls-files --exclude-standard --ignored --others`
- display the length of ignored files and only the 10 first (all the list in DEBUG)
- optimize the search of file in the ignored list

tpansino:
Hey @nvuillam, this PR is from @TimPansino, not me.

He is actually my brother though 😆

TimPansino:
Ah, didn't notice the PR from @davidB, looks about the same with more changes. I'll close this and add some comments over there.

nvuillam:
@tpansino @TimPansino ypu really are brothers ? 😁 
Coding in family must be nice 🍻

PR #689, opened 2021-08-24
Add cli_lint_mode to golangci-lint

NeilGerring:
Currently, the golangci-lint linter will validate each go file individually, this causes a problem if the file is using a function or variable declared in a different file within the same package. To solve this, I've added the cli_lint_mode in the linter description and updated the tests slightly. 

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/.automation/test/golang/golang_bad_01.go
+++ b/.automation/test/golang/golang_bad_01.go
@@ -1,3 +1,7 @@
+package main
+
 if len(in) == 0 {
-  return "", fmt.Errorf("Input is empty")
+	return "", fmt.Errorf("Input is empty")
 }
+
+sum(1, 2)
--- a/.automation/test/golang/golang_good_01.go
+++ b/.automation/test/golang/golang_good_01.go
@@ -4,4 +4,5 @@ import "fmt"
 
 func main() {
 	fmt.Println("hello world")
+	sum(1, 2)
 }
--- a/.automation/test/golang/golang_good_maths_01.go
+++ b/.automation/test/golang/golang_good_maths_01.go
@@ -1,0 +1,5 @@
+package main
+
+func sum(a int, b int) int {
+	return a + b
+}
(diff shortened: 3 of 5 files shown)


nvuillam:
That seems much better for performances, thanks :)

Let's see what the CI says, but i think it will be ok :)

NeilGerring:
Thanks for getting back so quick, just giving this another kick locally and it looks like it's not quite working as I expect. I'll stick a do not merge label on this for now and try fix it as soon as possible

nvuillam:
No need for specific label, you can define the PR as draft until you think it's ready to be reviewed :)

https://docs.github.com/en/github/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request#converting-a-pull-request-to-a-draft

NeilGerring:
This has become more of a pain in the backside then I had originally thought. 

If we specify the list of files and have multiple sub directories, the following error from golangci-lint will be returned:
`named files must all be in one directory; have ./ and ./foo/`

The following will work without any worries about the sub directories if the root directory has a go.mod and none of the others do: 
`golangci-lint run`

I think we'd need to search for all directories with a go.mod and run `golangci-lint run` in each. 

Any thoughts?

nvuillam:
There is the `project` cli_lint_mode that calls the linter on the whole directory (without referring to any file in particular), but  FILTER_REGEX_INCLUDE & FILTER_REGEX_EXCLUDE are not taken in account ( ignore must be managed on the linter config level)

If performances are better like that, that could be a solution

To browse the folders and call the linter on each folder containing go.mod , generic Mega-Linter architecture does not fit, but a [custom python class for this linter may do the job, like it does for other linters](https://github.com/nvuillam/mega-linter/tree/master/megalinter/linters) :)

nvuillam:
Seems CI is not happy ^^

PR #717, opened 2021-09-02
GitLab CI: Call mega-linter from Script

### LINKED ISSUE #716: How does this work in gitlab ci?
So i configured the project using "npx mega-linter-runner --install" for the documentation flavor -  I want to lint markdown files.

.gitlab-ci.yml:
```
mega-linter:
  stage: quality-gate
  image: nvuillam/mega-linter-documentation:v4
  script: ["true"]
  variables:
    DEFAULT_WORKSPACE: $CI_PROJECT_DIR
    DEFAULT_BRANCH: master
  artifacts:
    when: always
    paths:
      - report
 ```

.mega-linter.yml:

```
# Configuration file for Mega-Linter
# See all available variables at https://nvuillam.github.io/mega-linter/configuration/ and in linters documentation

APPLY_FIXES: all # all, none, or list of linter keys
DEFAULT_BRANCH: master # Usually master or main
# ENABLE: # If you use ENABLE variable, all other languages/formats/tooling-formats will be disabled by default
# ENABLE_LINTERS: # If you use ENABLE_LINTERS variable, all other linters will be disabled by default
DISABLE:
  # - COPYPASTE # Uncomment to enable checks of abusive copy-pastes
  - SPELL # Comment to disable checks of spelling mistakes # - SPELL # Uncomment to disable checks of spelling mistakes
SHOW_ELAPSED_TIME: true
FILEIO_REPORTER: false
```

However, I get the output:

```
Executing "step_script" stage of the job script
00:01
$ true         
Uploading artifacts...
WARNING: report: no matching files                 
ERROR: No files to upload                          
Cleaning up file based variables
00:00
Job succeeded
```

@nvuillam   What am I doing wrong?

Thank you

brunomartinspro:
Fixes #716 

## Proposed Changes

1. Changed the Script "Script" section to call megalinter.run
2. Changed the Documentation and templates

### Author/Contributor

### Reviewing Maintainer


bittner:
Can you explain the value of your change?

I do agree that `script: 'true'` is odd and not particularly self-explanatory in GitLab CI. But odds are that developers will easily introduce a typo in `python -m megalinter.run`. There must be a simpler solution, a better one. Maybe Mega-Linter could install a Python [entrypoint script](https://stackoverflow.com/questions/774824/explain-python-entry-points), e.g. `mega-linter`, and we should suggest to call that?

Also, if we introduce such an "API" it would make sense to make this more open. The `mega-linter` CLI may allow to list all available linters and call them selectively using command line options and arguments.

_I understand that my suggestion defers a "quick solution" and I apologize for that. This was not my original intention._

nvuillam:
@bittner that would be really nice to have an extended Mega-Linter CLI, but for the moment mega-linter-runner can only install mega-linter config files, and run the docker image

As running a big docker image locally is heavy, i'm not sure implementing such option would deserve all the invested time it would need :/

bittner:
> As running a big docker image locally is heavy, i'm not sure implementing such option would deserve all the invested time it would need :/

For the start, creating a Python entrypoint would be a matter of minutes. I would simply add a few lines to [setup.py](https://github.com/nvuillam/mega-linter/blob/master/megalinter/setup.py#L34) ... assuming that you install `megalinter` as a Python package (not just copy the source code) in the Docker image. :thinking: 

Should I go ahead and create a PR?

nvuillam:
If it's a matter of minutes and you can guarantee no regression, go for the PR ! :)

bittner:
> If it's a matter of minutes and you can guarantee no regression, go for the PR ! :)

I created PR #776, and as usual there is no guarantee for anything. It's a free software license. :roll_eyes: :smiley:

nvuillam:
> I created PR #776, and as usual there is no guarantee for anything. It's a free software license. 🙄 😃

mega-linter-runner created mega-linter alias in addition to mega-linter-runner, I removed it :)

I have no idea what you want to do with mega-linter as python alias, but I'm curious to see that :)

bittner:
> I have no idea what you want to do with mega-linter as python alias, but I'm curious to see that :)

Well, this is easy to explain with this PR: Instead of `python -m megalinter.run` we should be able to simply call `mega-linter`, which is an obvious (and self-explanatory) choice for invoking the tool.

What is the role of the (JavaScript-driven) `mega-linter-runner` CLI command? This is a bit opaque. (Why is there a JavaScript implementation for invoking what is implemented in Python? I see that some Docker is involved, but I don't yet understand the entire strategy.)

nvuillam:
The npm package is used for:
- installation (yeoman generator)
- running locally the docker image

Could have been a python package too but as i'm familiar with node I did it in node.js ^^ (and it can be called without install using npx )

PR #760, opened 2021-09-21
Fix PR 604

### LINKED ISSUE #604: Detached head state in git after running MegaLinter
**Describe the bug**
Running megalinter locally causes repo to end up in detached head state.

**To Reproduce**

1. `git status` -> `On branch feature-linting`
1. `mega-linter -f python --fix` -> Run to completion
1. `git status` -> `HEAD detached at dd492c19c`

**Expected behavior**
After running, repo should be on the same branch it started on.

**Screenshots**
N/A

**Additional context**
Megalinter logs show a warning, possibly related.

```
Listing updated files in [/tmp/lint] using git diff, then filter with:
Warning: Unable to pull current branch dd492c19c92ff5d4c74ae5659e36d82c45ef4eb8
Git diff :
.github/workflows/mega-linter.yml
.gitignore
.mega-linter.yml
pyproject.toml
setup.cfg
tests/framework_graphene/_target_application.py
```

My config file

```
APPLY_FIXES: none # all, none, or list of linter keys
DEFAULT_BRANCH: main # Usually master or main
SHOW_ELAPSED_TIME: true
FILEIO_REPORTER: false
PRINT_ALPACA: false
VALIDATE_ALL_CODEBASE: false
ENABLE_LINTERS: # If you use ENABLE_LINTERS variable, all other linters will be disabled by default
  - MARKDOWN_MARKDOWN_LINK_CHECK
  - MARKDOWN_MARKDOWNLINT
  - PYTHON_BANDIT
  - PYTHON_BLACK
  - PYTHON_FLAKE8
  - PYTHON_ISORT
  - PYTHON_PYLINT
  - RST_RST_LINT
  - RST_RSTCHECK
  - RST_RSTFMT
  - YAML_PRETTIER
  - YAML_V8R
  - YAML_YAMLLINT
```

Full logs attached

[mega-linter.log](https://github.com/nvuillam/mega-linter/files/6909337/mega-linter.log)

I also noticed when I run it with `IGNORE_GITIGNORED_FILES: true` I get a different error as well, could also be related. Looks like an iterable was expected here but a string was passed.

```
Warning: Unable to pull current branch 5468884b48d22c777537da5391ed835ad1b11bc3
Git diff :
.github/workflows/mega-linter.yml
.gitignore
.mega-linter.yml
pyproject.toml
setup.cfg
tests/framework_graphene/_target_application.py
- File extensions: .md, .py, .rst, .yaml, .yml
Unable to list git ignored files (Cmd('g') not found due to: FileNotFoundError('[Errno 2] No such file or directory: 'git status --ignored'')
  cmdline: g i t   s t a t u s   - - i g n o r e d)
```

noskovao:
Fixes #604

## Proposed Changes

1. Remove unnecessary checkout inside a code which causes detached head after execution

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/megalinter/MegaLinter.py
+++ b/megalinter/MegaLinter.py
@@ -475,25 +475,16 @@ class Megalinter:
         logging.info(
             "Listing updated files in ["
             + self.github_workspace
-            + "] using git diff, then filter with:"
+            + "] using git diff."
         )
         repo = git.Repo(os.path.realpath(self.github_workspace))
         default_branch = config.get("DEFAULT_BRANCH", "master")
-        current_branch = config.get("GITHUB_SHA", "")
-        if current_branch == "":
-            current_branch = repo.active_branch.commit.hexsha
-        try:
-            repo.git.pull()
-        except git.GitCommandError:
-            try:
-                repo.git.checkout(current_branch)
-                repo.git.pull()
-            except git.GitCommandError:
-                logging.info(f"Warning: Unable to pull current branch {current_branch}")
-        repo.git.checkout(default_branch)
-        diff = repo.git.diff(f"{default_branch}...{current_branch}", name_only=True)
-        repo.git.checkout(current_branch)
-        logging.info(f"Git diff :\n{diff}")
+        current_commit_sha = config.get("GITHUB_SHA", "")
+        if current_commit_sha == "":
+            current_commit_sha = repo.active_branch.commit.hexsha
+        repo.git.fetch()
+        diff = repo.git.diff(f"origin/{default_branch}...{current_commit_sha}", name_only=True)
+        logging.info(f"Modified files:\n{diff}")
         all_files = list()
         for diff_line in diff.splitlines():
             if os.path.isfile(self.workspace + os.path.sep + diff_line):


nvuillam:
The diff should not be compared to the latest commit but with the target branch of the PR, so your update will probably break the case when Mega-Linter is called with VALIDE_ALL_CODE_BASE = false in a PR CI ^^

But... maybe we could manage different code called depending if we have CI variable set or not ? ( no CI variables means Mega-Linter runs locally )

noskovao:
Ok, agree, but IMO I think we do not need to perform any actions with the git repo during the execution of Mega-Linter (checkout, pull, etc). If it is necessary, it should be done prior to linter execution.
I've uploaded another commit where we are doing a simple git fetch just in case.
Moreover, in this part of code, I do not see the necessity of "git checkout" because anyway, we are getting back to ${current_branch} :)

nvuillam:
As Mega-Linter is a fork of Super-Linter, which as been build by Github DevOps teams, I assume they are more git experts than me (spoiler: i am absolutely no git expert ^^) , and the current python code is a mirror of [what they did in bash](https://github.com/github/super-linter/blob/master/lib/functions/buildFileList.sh) (or what it used to be)

What you propose seems legit, I'm just afraid of regressions (mostly by lack of knownledge)

Are you really sure that there will be no regression if I merge this PR ? :)

noskovao:
I believe there shouldn't be a regression because currently, it is doing checkout twice, first to "default_branch" commit, then getting back to "current_branch" commit and become detached.
BTW notice we need to specify remote for DEFAULT_BRANCH, not only branch name. Otherwise, it is not working :(
So I've added this in the code, lets see how the checks go.

noskovao:
@nvuillam ok, now it is ready to be merged.

nvuillam (approved):
Thanks for your contribution :)
CI seems happy... let's have a leap of faith ^^

Please check with insiders/latest version Mega-Linter so we'll be sure to not break anything in the next official release :)

Kurt-von-Laven:
Love it; this change is definitely an improvement, and I don't foresee any regression coming from this. See #850 for a related pre-existing issue.

PR #770, opened 2021-09-24
Implementation of solidity linter, added the descriptor and the test …

leogalbu:
## Proposed Changes

1. Added slither, solidity linter

### Author/Contributor

### Reviewing Maintainer


nvuillam:
CI not happy ^^
You can run the test class locally in VsCode on PyCharm, it will avoid you to wait for the (long) CI job to know where is the issue

leogalbu:
this is great, thank you very much :)

leogalbu:
@nvuillam I followed your advice and tested locally, replicated the error and fixed it by setting solc in the environmental variables.
Do you have any advice on how to set this on the config file?

nvuillam:
What is solc ?
If you need an env var to be set, you can add that in install descriptor property (dockerfile)

leogalbu:
Solc is the solidity compiler, it is required to use slither, the  static analysis framework.

PR #771, opened 2021-09-24
Remove unnecessary jq due to CVE-2016-407

noskovao:
Clair check reporting that jq contains critical CVE-2016-407
https://nvd.nist.gov/vuln/detail/CVE-2016-4074
This might be false-positive https://github.com/quay/clair/issues/852
but anyway I didn't find anything which is using jq in the code,
I assume it is leftover from the previous code.


nvuillam (approved):

PR #776, opened 2021-09-25
Create CLI command when installing megalinter Python package

bittner:
Adds a Python entry point console script, which allows to intuitively invoke the CLI.

Relates to #717.

### Author/Contributor

### Reviewing Maintainer


nvuillam:
The error i see in CI is not your fault, it seems that tsqllint decides sometimes to not return help or version >8<

PR #838, opened 2021-10-11
Update suggestion typo

ChrisChinchilla:
The grammar for letting mega linter suggest a language was not quite correct, so tweaked


ChrisChinchilla:
@nvuillam NP, I noticed a few more places I might get around to checking some day

PR #855, opened 2021-10-15
Prettify console output

llaville:
A minor PR that allow to fix some formatting console output (when DEBUG mode is activated):

Currently we can see some bullet list that well not handle the first item of a file list. See following 

format_bullet_list
And also could provide an empty list with a single dash line when no file are kept
empty_bullet_list

Their is also file list to kept, see 
linter_filelist_bullet

That may us avoid to count files with someting like 
linter_filelist_bullet_after

## Proposed Changes

1. Add a new function `format_bullet_list` in `utils.py`
2. Replaces all hardcoded attempt to format bullet list with calls to this function

### Author/Contributor

### Reviewing Maintainer


nvuillam:
Great, thanks :)

PR #861, opened 2021-10-17
Add sqlfluff-templater-dbt plugin to sqlfluff installation instructions

### LINKED ISSUE #842: Add optional DBT package for Mega-Linter for SQLFluff from v0.7.0
Hi, I'm one of the maintainers of SQLFluff.

From our next release (0.7.0) we are splitting our some of the functionality into a separate, optional plugin that users can install only if they need that functionality.

The reason it's being split out is not all users need this, and it allows a separate release cadence, and to clean up our code, and to reduce the need for users who don't need it to install this code. There is no harm to install it even if it's not used.

Therefore I think the Mega-Linter should include this by default. It won't be used unless referred to in the user's SQLFluff config file, but would need to be included in your Docker builds to allow them to continue using this functionality (which they may well have been happily using in the current version of the Mega Linter). 

I've raised https://github.com/sqlfluff/sqlfluff/issues/1600 in the SQLFluff repo, as a reference for this to track adding this here so feel to ask any questions here or there.

The plugin is called `sqlfluff-templater-dbt` is currently part of the main SQLFluff repo, but is a separate pip package (the package is already available now as an alpha release (0.7.0a4): https://pypi.org/project/sqlfluff-templater-dbt/).

As it's an optional dependency it's not included in our Python `requirements.txt` dependence file so won't be installed by default based on how SQLFluff is currently pulled into the Mega Linter. How can we get this plugin added in as a dependency to the Mega-Linter at the same time as SQLFluff to ensure users can continue to avail of that functionality if they want it? As I said above, those who do not use DBT will not have any issues with this package being installed.

Happy to open a PR for this if you could give me some pointers of where to add it. Note this should only be added when the Mega-Linter upgrades to SQLFluff 0.7.0 which we expect to send live in next few days.

nvuillam:
Fixes #842

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Note: Can be used with `nvuillam/mega-linter@insiders` in your GitHub Action meg
 - Upgrades
   - Base docker image python:3.9.6-alpine3.13 to python:3.9.7-alpine3.13
   - Automerge internal job pascalgn/automerge-action-0.14.2 to pascalgn/automerge-action-0.14.3
+  - Add sqlfluff-templater-dbt plugin to sqlfluff installation instructions
 - Config reporter: Parse `.vscode/extensions.json` as json5 (with comments)
 - Add eslint-plugin-jsx-a11y dependency
 - Rename default PHPStan config file, from `phpstan.neon` to `phpstan.neon.dist` accordingly to [PHPStan resolution priority](https://phpstan.org/config-reference#config-file)
--- a/Dockerfile
+++ b/Dockerfile
@@ -146,6 +146,7 @@ RUN pip3 install --no-cache-dir --upgrade \
           'snakemake' \
           'snakefmt' \
           'sqlfluff' \
+          'sqlfluff-templater-dbt' \
           'checkov>=2.0.269' \
           'yamllint'
 #PIP__END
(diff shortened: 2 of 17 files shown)


nvuillam:
@tunetheweb is this PR still necessary ?

tunetheweb:
I'm not sure to be honest!

On the one hand, this was functionality that was available in SQLFluff prior to 0.7.0 hence why I suggested adding it.

On the other hand, I've not heard of any complaints on this or the Super Linter and that was released a while back.

The dbt that this functionality adds back in due to it being an optional dependency, is a Python package - as is SQLFluff - so it could be most dbt users use it directly, rather than through the Mega or Super Linter. In which case this is not needed.

Maybe close for now, to keep the package and dependencies down, and reopen if people ask for it?

nvuillam:
@tunetheweb I confirm tht CI that calls sqlfluff still works ^^

thanks for your honest answer, let's follow your suggestion :)

jakeberesford-palmetto:
I would have liked to use this feature. I understand not wanting to bloat the dependencies, and actually just adding the dbt templater wouldn't be enough, since [the templater also requires the relevant dbt-adapter be installed](https://docs.sqlfluff.com/en/stable/configuration.html#installation-configuration)

Is it possible to add a configuration option that would allow users to specify additional packages to be installed?

For now, I can implement a SQLFluff CI action, but I would prefer all linting to be handled by megalinter (I'm using it for markdown and yaml files in my dbt project)

Kurt-von-Laven:
[SQL_SQLFLUFF_PRE_COMMANDS](https://megalinter.github.io/latest/descriptors/sql_sqlfluff/) should work for this purpose, but please let us know if you run into any issues.

PR #867, opened 2021-10-18
Introduces class constants to avoid hardcoded paths

llaville:
Fix extra slash heading in front of rules config file in console reporter

extra_slash

## Proposed Changes

1. Add class constants: 
    - `TEMPLATES_DIR` to replace hardcoded path `/action/lib/.automation/`
    - `DEFAULT_WORKSPACE_DIR` to replace hardcoded path `/tmp/lint/`
2. Be carefull to always have a trailing slash

**WARNING** all hardcoded values were not replaced. Perharps we could do ! We fix here only the extra slash in front of rules config file identified

### Author/Contributor

### Reviewing Maintainer


nvuillam (approved):
nice, thanks :)

PR #897, opened 2021-10-29
Only invoke git fetch when default branch missing, and only fetch the default branch if so.

### LINKED ISSUE #896: Mega-Linter Runner Crashes In Incremental Mode Locally
**Describe the bug**
The `git fetch` on the `VALIDATE_ALL_CODEBASE=false` path results in a crash when run locally.

```
stderr: Host key verification failed.
fatal: Could not read from remote repository.
```

**To Reproduce**

1. `cd` into a Git repository.
2. Run `git remote set-url origin <ssh_url>` if you aren't already using ssh.
3. Run Mega-Linter Runner locally using `latest` Docker image with `VALIDATE_ALL_CODEBASE=false`.

**Expected behavior**
I expected Mega-Linter Runner to run successfully.

**Additional context**
This is a long-standing pre-existing issue that was exposed by the removal of a preceding crash in #851.

Kurt-von-Laven:
Fixes #896.

## Proposed Changes

1. When running in incremental mode locally on a private repository, invoking `git fetch` fails because the Docker image doesn't have access to your personal SSH keys (or a TTY via which to enter your username and password). Running `git fetch` also prevents Mega-Linter from running offline, and weakens the protection offered by `git push --force-with-lease`. However, the fetch is needed when running in incremental mode on a clone that lacks `origin/<default_branch>`. Notably, v2+ of the GitHub Actions checkout action defaults to shallow clones of depth 1, which always lack `origin/<default_branch>`. The end user is responsible for ensuring that their clone has `origin/<default_branch>` when running in incremental mode locally.
2. Instruct git to only fetch the default branch. Improve performance of incremental mode when using clones that lack `origin/<default_branch>`. There is no need to fetch every single branch when we only compare against the default branch.

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/megalinter/MegaLinter.py
+++ b/megalinter/MegaLinter.py
@@ -486,7 +486,10 @@ class Megalinter:
         default_branch_remote = f"origin/{default_branch}"
         if default_branch_remote not in [ref.name for ref in repo.refs]:
             # Try to fetch default_branch from origin, because it isn't cached locally.
-            repo.git.fetch()
+            repo.git.fetch(
+                "origin",
+                f"refs/heads/{default_branch}:refs/remotes/{default_branch_remote}"
+            )
         diff = repo.git.diff(default_branch_remote, name_only=True)
         logging.info(f"Modified files:\n{diff}")
         all_files = list()
--- a/megalinter/MegaLinter.py
+++ b/megalinter/MegaLinter.py
@@ -483,8 +483,11 @@ class Megalinter:
         )
         repo = git.Repo(os.path.realpath(self.github_workspace))
         default_branch = config.get("DEFAULT_BRANCH", "master")
-        repo.git.fetch()
-        diff = repo.git.diff(f"origin/{default_branch}", name_only=True)
+        default_branch_remote = f"origin/{default_branch}"
+        if default_branch_remote not in [ref.name for ref in repo.refs]:
+            # Try to fetch default_branch from origin, because it isn't cached locally.
+            repo.git.fetch()
+        diff = repo.git.diff(default_branch_remote, name_only=True)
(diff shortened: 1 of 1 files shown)


nvuillam:
@Kurt-von-Laven when we are in CI mode , i think it can be useful to check the other branches to be able to make diff :)

So... maybe do that only if isCI is False ?

Or even an env variable MEGALINTER_GIT_FETCH with default true, that you could override with no for your local cases ? :)

Kurt-von-Laven:
Yes, apologies for the inadequately tested PR. Here is a revised version that works both locally and in CI. I haven't yet thought of a sane way to support local shallow clones in incremental mode, and I'm not entirely certain that supporting them would be a good thing overall since (a) it would probably imply that Mega-Linter has access to their personal SSH keys and (b) the user can simply run `git fetch` (or the more specific `git fetch origin refs/heads/main:refs/remotes/origin/main` for optimal performance) themselves beforehand. In any case, that is a pre-existing issue as opposed to a regression.

nvuillam:
Let's go ;)
For your next PR, please think to update also CHANGELOG.md, that's a pain in the a** to do it manually later ^^

Kurt-von-Laven:
What is your sentiment regarding CHANGELOG automation tools (e.g., [commitizen](https://commitizen-tools.github.io/commitizen/))?

nvuillam:
@Kurt-von-Laven I heard of them, but for me there is a difference between tech details explained in commits, and documentation for humans, that's why I prefer to let humans write the docs :)

Kurt-von-Laven:
Yes, that does seem more valuable to our users than an automated CHANGELOG would be!

PR #921, opened 2021-11-02
Use GH actions concurrency to cancel runs

### LINKED ISSUE #920: Use concurrency option of GH actions to cancel previous workflow runs
This is a follow-up to #622, i.e. about cancelling previous workflow runs when a new push to a branch occurs.

GitHub actions directly offers a means to prevent [concurrent workflow runs](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency) and [concurrent job runs](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idconcurrency) and to cancel any previous runs.

**Advantages:**
* It's not another job, so
  * cancellation happens faster,
  * it does not show up as a check in PRs,
  * you don't get billed for it.
* Not entirely sure about this, but I think you don't get notified (e.g. by mail) about cancelled runs.

**Disadvantages:**
* [skip-duplicate-actions](https://github.com/fkirc/skip-duplicate-actions) offers more options.
* One needs to be careful that only the mega-linter workflow is cancelled by assigning a unique concurrency group.

**Example:**

```yml
name: Mega-Linter

on:
  pull_request:
    branches: [main]

# Workflow can only run once per branch and any previous pending or in-progress runs are cancelled.
concurrency:
  group: ${{ github.ref }}-${{ github.workflow }}
  cancel-in-progress: true

env:
  APPLY_FIXES: all
  APPLY_FIXES_EVENT: pull_request
  APPLY_FIXES_MODE: commit
  FILTER_REGEX_EXCLUDE: (\.github/workflows/)

jobs:
  lint:
    # As usual...
```

Keep everything as is, which is also perfectly acceptable (so feel free to close this).

**Additional context**
Links (same as above):
- [Workflow concurrency](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency)
- [Job concurrency](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idconcurrency)

lars-reimann:
Fixes #920

## Proposed Changes

1. Use [workflow concurrency](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency) instead of [skip-duplicate-actions](https://github.com/fkirc/skip-duplicate-actions)
2. Update `mega-linter-runner` template
3. Add a regex replacement to the updater ([job removal](https://regex101.com/r/kACYMf/1/) + [comment removal](https://regex101.com/r/IGPfs5/1))
  
Capabilities of the regex:
  * Works with varying whitespace (including indentation)
  * Keeps any jobs that come before or after the `cancel_duplicates` job.

Downsides of the regex:
  * It will not work if the user has renamed the `cancel_duplicates` job. I don't see a way to reliably fix this, however.
  * It removes comments between the end of the `cancel_duplicates` job and the start of the next. This can be fixed by matching the end of the `cancel_duplicates` job more exactly, but that again makes it more likely that the automatic update fails if the user has changed the workflow.

### Author/Contributor

### Reviewing Maintainer


nvuillam (approved):
Amazing regex !
And you even corrected some bug I let while migrating to V5, thanks a lot !
Please can you also add a line in CHANGELOG.md ?

lars-reimann:
@nvuillam I guess an additional changelog entry is not needed anymore, right?

nvuillam:
@lars-reimann you did such a fine job that i thought I owed you that ;)

PR #925, opened 2021-11-03
Fix upgrader removing all jobs after cancel_duplicates but last

lars-reimann:
Fixes #

The regular expression added in #921 was a little too greedy and removed all jobs in a workflow starting with `cancel_duplicates` but the last.

**Example:**

```yml
jobs:
  preceding_job:
    #..
  # Cancel duplicate jobs: https://github.com/fkirc/skip-duplicate-actions#option-3-cancellation-only
  cancel_duplicates:
    name: Cancel duplicate jobs
    runs-on: ubuntu-latest
    steps:
      - uses: fkirc/skip-duplicate-actions@master
        with:
          github_token: ${{ secrets.PAT || secrets.GITHUB_TOKEN }}
          cancel_others: true

  intermediate_job:
    #..
  build:
    #..
```

would become

```yml
jobs:
  preceding_job:
    #..
  # Cancel duplicate jobs: https://github.com/fkirc/skip-duplicate-actions#option-3-cancellation-only
  build:
```

**missing the intermediate_job**.

## Proposed Changes

1. Fix regex (see [regex101](https://regex101.com/r/Vand2F/1))
2. Update test case

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Note: Can be used with `megalinter/megalinter@beta` in your GitHub Action mega-l
   - [cfn-lint](https://github.com/martysweet/cfn-lint) from 0.54.4 to **0.55.0** on 2021-11-03
   - [phpstan](https://phpstan.org/) from 1.0.0 to **1.0.1** on 2021-11-03
 <!-- linter-versions-end -->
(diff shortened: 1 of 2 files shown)


nvuillam (approved):
Great, thanks for the fix !

PR #957, opened 2021-11-08
Allow GithubStatusReporter to work for other CI platforms

RGPosadas:
Fixes N/A

## Proposed Changes

According to the [docs](https://github.com/megalinter/megalinter/blob/main/docs/installation.md#github-action), megalinter should be able to mark the status of each linter in the Checks section of a PR. While this works as stated for GithubCommentReporter, it does not for GithubStatusReporter as it expects a `GITHUB_RUN_ID` in order to activate the reporter. If we are not using GitHub Actions, then this reporter will not work for other CI tools.

1.  Remove mandatory check for `GITHUB_RUN_ID`. This will always exist if it comes from a GitHub Action.
2. If not a GitHub Action, then `target_url` can be retrieved from an env var `GITHUB_TARGET_URL`.

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Note: Can be used with `megalinter/megalinter@beta` in your GitHub Action mega-l
 - Documentation versioning with mike
 - Accordingly, to official [PHPStan documentation](https://phpstan.org/user-guide/rule-levels), the TEMPLATES/phpstan.neon.dist config file set default level to zero.
 - Downgrade dotnet from 6.0 to 5.0, to be compliant with tsqllint
+- Allow GithubStatusReporter to work for other CI platforms
 
 - Linter versions upgrades
   - [cspell](https://github.com/streetsidesoftware/cspell/tree/master/packages/cspell) from 5.12.5 to **5.12.6** on 2021-11-04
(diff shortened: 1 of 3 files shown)


nvuillam:
@RGPosadas that's a great idea, but just changing the target url won't be enough, as the message is formatted for Github Actions (link to Action artifacts, use of github related env vars...

I think the good architecture is the following:
- Create a parent class scm_reporter.py
- Move everything generic to all CI tools in from GithubCommentReprter to ci_reporter_root.py, like the buid of the message
- Make GithubCommentReporter inherit of ci_reporter_root and keep in it only github related code
- Create a class GitLabCommentReporter.py, inheriting of ci_reporter_root, with only gitlab related code
- Create a class AzureCommentReporter.py, inheriting of ci_reporter_root, with only Azure Pipeline related code
- etc...

RGPosadas:
Hi @nvuillam, 
Ah, makes sense! However, if we were to keep it generic to GitHub as the SCM for now, then it would work as I got this working with my Jenkins. 

The only thing I had to manipulate was the `target_url` for it to not expect a GitHub Actions workflow, but rather a redirect to my Jenkins build output. To enable the checks to be shown as part of the PR checklist, the actual request URL being sent to is `url = f {github_api_url}/repos/{github_repo}/statuses/{sha}"` which doesn't need GitHub Actions. I still see this PR as a viable option.

However, should you want to continue with architecture you mentioned above with multiple SCMs, I can close this PR and create an issue for this :)

nvuillam:
@RGPosadas do you mean that with your update, Github + Jenkins will work ?
If yes, I can merge it :)

Please can you just put your changelog update in [Unlereased] section ? 
I'll manage myself the release in an official version ^^

RGPosadas:
@nvuilla Yup they both work :) 
Here's an example of the status checks running with some information redacted. Clicking on any of the `Details` link will redirect the user to the `GITHUB_TARGET_URL` set.

Screen Shot 2021-11-10 at 11 04 20 AM

nvuillam (approved):
Great, nice job :)

PR #968, opened 2021-11-09
Use root user in the Jenkins pipeline

McFy49:
Jenkins fails with "process never started.." error:

It run ok locally but won't work without root privilege on a Jenkins pipeline

Fixes #

## Proposed Changes

1. ...
2. ...
3. ...

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/README.md
+++ b/README.md
@@ -445,7 +445,7 @@ stage('Mega-Linter') {
     agent {
         docker {
             image 'megalinter/megalinter:v5'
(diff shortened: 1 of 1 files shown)


nvuillam (approved):
Nice catch, thanks !

McFy49:
> Nice catch, thanks !

Merci ✌️

PR #1137, opened 2021-12-30
introduce shell script to convenience running from ssh

InbarRose:
Fixes #

## Proposed Changes

1. introduce shell script to allow running commands simply through ssh using tmux

### Author/Contributor

### Reviewing Maintainer


nvuillam:
@InbarRose you should check in your mails, it would be easier if you accept the invitation to be a contributor :) ( no more forks, + CI allowing you to directly generate testing docker images after your updpates :) )

PR #1182, opened 2022-01-19
Make python linters play nice with each other

schorlton:
I may be missing something, but putting 4+ python linters together seems like a good challenge to make them all play nice. Fortunately, Black as a guide to do so: [https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html](https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html). I chose Black as it's the most popular Python linter currently implemented by stars on Github.

I didn't see these settings set for Pylint and Flake8 but did see the profile set for `isort`. The Pylint config had a lot of stuff in there so may have missed something. I also note the Black suggested rules to disable in Pylint are named differently than the current rules to disable in Pylint, but I tried out Pylint with this config and seems to work.

Also fixed this very minor issue: https://github.com/megalinter/megalinter/pull/951#discussion_r744781586

## Proposed Changes

1. Change Pylint and Flake8 defaults to be compatible with Black

### Author/Contributor
- [ x] If documentation is needed for this change, has that been included in this pull request

### Reviewing Maintainer

### DIFF
--- a/TEMPLATES/.flake8
+++ b/TEMPLATES/.flake8
@@ -1,2 +1,3 @@
 [flake8]
-max-line-length = 120
+max-line-length = 88
+extend-ignore = E203
--- a/TEMPLATES/.python-lint
+++ b/TEMPLATES/.python-lint
@@ -128,7 +128,9 @@ disable=print-statement,
         dict-items-not-iterating,
         dict-keys-not-iterating,
         dict-values-not-iterating,
-        import-error
+        import-error,
+        C0330,
+        C0326
 
 # Enable the message, report, category or checker with the given id(s). You can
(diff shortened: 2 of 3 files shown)


schorlton:
@nvuillam if you point me to how to fix the failing check, I can cut another commit. Thanks!

nvuillam:
@schorlton all linters have test cases run during CI

As you changed the default flake8 config, the flake8 test case do not work anymore ^^

So to solve the issue, you must make that flake8 does not trigger an error when runned on this file ->
https://github.com/megalinter/megalinter/blob/main/.automation/test/python/python_good_1.py :)

schorlton:
Awesome, should be working now :)

PR #1209, opened 2022-01-26
Add yamllint compatibility for prettier

schorlton:
Fixes incompatibility between prettier and yamllint for yaml files. Prettier has a standard for 1 space between content and comment in yaml, and it fixes files to this standard. They do not plan on changing it or enabling the user to change it: https://github.com/prettier/prettier/pull/10926

yamllint enables configuation, so let's just go with the prettier (and more popular/relaxed) standard.

## Proposed Changes

Also added in a test case that shouldn't fail on a singly spaced comment.

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/.automation/test/yaml/yml_good_1.yml
+++ b/.automation/test/yaml/yml_good_1.yml
@@ -13,6 +13,6 @@ env:
   es6: true
   jest: true
 
-Here: there
+Here: there # good comment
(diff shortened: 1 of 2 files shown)


nvuillam:
Great, thanks for your contribution :)

PR #1261, opened 2022-02-09
Fix suggestion to use proper variable name

benoitbleuze:
Fixes #

The suggestion to use a token on gitlab to publish comments is not in sync with the actual variable.
This PR fixes the text to match the code.

## Proposed Changes

1. rename the variable in the comment.

### Author/Contributor

### Reviewing Maintainer


nvuillam (approved):
Good catch, thanks :)

PR #1356, opened 2022-03-26
Github Comment Reporter: Use a html comment to mark the comment

### LINKED ISSUE #1355: GitHub Comment reporter fails to find existing comment when using a CI other than GitHub
The GitHub Comment reporter identifies a pre-existing Mega-Linter comment in a PR by looking for a specific line of text:

https://github.com/megalinter/megalinter/blob/abfb0dccab7ff24e8ac4363a1616df83babf97a3/megalinter/reporters/GithubCommentReporter.py#L88-L93

However, that line of text is only going to be there if the comment was posted *from a GitHub action runner*, because only then a non-empty `action_run_url` value has been passed to `build_markdown_summary()`. If that variable is an empty string, the text that is included is different:

https://github.com/megalinter/megalinter/blob/abfb0dccab7ff24e8ac4363a1616df83babf97a3/megalinter/utils_reporter.py#L88-L94

We are using Jenkins (where we provide enough context and a token for the reporter to run successfully), but each PR update triggers a new comment.

The check seems to be very fragile in any case. Perhaps the reporter could include a HTML comment with a know identifier, instead? E.g. append `”\n\n<!— mega_linter_github_comment_reporter comment marker —>` to the generated markdown and look for that text instead?

The marker could even incorporate the workflow name (`GITHUB_WORKFLOW`) and jobid (`GITHUB_JOB`), if set, to make it possible to run more than one MegaLinter job per PR.

mjpieters:
Fixes #1355

This replaces the old method, which looked for the "see error details" link,
but that link was not always present in comments.

The comment records the current workflow and jobid, so multiple mega-linters
can comment on the same PR without colliding.

### Author/Contributor

### Reviewing Maintainer


nvuillam:
@mjpieters thanks for your contribution !
the issue is not related to your updates, I'll fix it in another PR, then upgrade your branch, then I'm pretty sure yours will pass :)

nvuillam (approved):

mjpieters:
> Quick question: Did you manage to upload report artifacts in jenkins ?
> if so, would it be possible to share your script to update [jenkins installation instructions](https://github.com/megalinter/megalinter#jenkins) ? ( I used jenkins several years ago but I do not have access to a jenkins instance these days)

I'm a bit of a Jenkins novice myself, but yes, artefacts are being uploaded. Adapting your example to add artefact archiving would look like this:

```groovy
stage('MegaLinter') {
    agent {
        docker {
            image 'megalinter/megalinter:v5'
            args "-u root -e VALIDATE_ALL_CODEBASE=true -v ${WORKSPACE}:/tmp/lint --entrypoint=''"
            reuseNode true
        }
    }
    steps {
        sh '/entrypoint.sh'
    }
    post {
        always {
            archiveArtifacts allowEmptyArchive: true, artifacts: 'mega-linter.log,report/**/*', defaultExcludes: false, followSymlinks: false  
        }
    }
}
```

By using `post { always { ... } }` the `archiveArtifacts` command runs whatever the outcome of the megalinter run.

nvuillam:
Many thanks for the tip, I'll update the doc :)

PR #1359, opened 2022-03-27
Improve the comment marker format

mjpieters:
This is a follow-up to #1356

Instead of just joining the workflow name and jobid without spacing,
encapsulate both in quotes and give them a `workflow=` and `jobid=` prefix to
aid debugging.

E.g. if you view the megalinter comment on a currently active PR you can see
that the workflow and jobid are currently concatenated:

```sh
$ gh pr view 1358 --json comments -q \
  '.comments[] | select(.body | contains("<!-- megalinter:")) | .body | split("\n")[-2]'
<!-- megalinter: github-comment-reporter MegaLinterbuild -->
```

It is unclear now where the workflow name ends and the jobid starts, or if one
of these two values isn't available.

This PR will change that to ``

### Author/Contributor

### Reviewing Maintainer


nvuillam (approved):
I agree it's even nicer like that, thanks :)

PR #1485, opened 2022-05-20
Run as non-root user after the image is built

MikolajBalcerek:
Before this change, megalinter's entrypoint would run as user 0:0.

```
docker run --entrypoint /bin/bash -it megalinter/megalinter:v5
bash-5.1# id -u
0
bash-5.1# id -g
0
```

This change makes it so default user is 1000:1000, which on most systems is non-root.
```
docker run --entrypoint /bin/bash -it mikolajbalcerek/megalinter:v5
bash-5.1$ id -u && id -g
1000
1000
```

Thus the downgraded `entrypoint.sh` won't be able to accidentally change files requiring root user on host system.

The workaround of running via: `docker run --user $(id -u):$(id -g)` stopped working in v5 after git config changes were introduced https://github.com/megalinter/megalinter/blob/main/entrypoint.sh#L7-L15

See this practice recommended in several sources:
- https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#non-root-user
- https://docs.bitnami.com/tutorials/why-non-root-containers-are-important-for-security

## Proposed Changes

1. ...
2. ...
3. ...

### Author/Contributor

### Reviewing Maintainer


Kurt-von-Laven:
I love the idea of changing this, but it requires a higher level of testing than most changes as it is likely to break many things. I suggest adding `adduser` after [this line](https://github.com/megalinter/megalinter/blob/c0d8c6e8cadd39c6901f61b854c62fab0854d316/.cspell.json#L294) to address the CSpell false positive. It looks like there may be some files that are owned by root that need to be cleaned up somehow to fix the permission errors?

Kurt-von-Laven:
Ideally on Linux folks would take advantage of Docker's rootless mode so that this is a non-issue, but I definitely agree that running as root is bad security practice. Here is [a GitHub action](https://github.com/ScribeMD/rootless-docker) that runs Docker in rootless mode for folks using GitHub Actions.

PR #1511, opened 2022-06-04
minor: update README's toc

eitsupi:


nvuillam (approved):
Thanks for catching and correcting :)

PR #1534, opened 2022-06-18
Add OUTPUT_FORMAT and OUTPUT_DETAIL to schema

bdovaz:
The following issue mentions the OUTPUT_DETAIL variable, which is not included in the schema file.

Issue where it is mentioned: https://github.com/megalinter/megalinter/issues/413

In addition to that, I also noticed that in the examples & workflows it is misused. In the code **OUTPUT_DETAIL** is used, but in the examples  & workflows **OUTPUT_DETAILS**:

Code: https://github.com/megalinter/megalinter/search?q=OUTPUT_DETAIL

Examples & Workflows: https://github.com/megalinter/megalinter/search?q=OUTPUT_DETAILS

And finally, when looking for **OUTPUT_DETAIL** references I see that there is also an undocumented **OUTPUT_FORMAT** variable with the possible values: tap, text. I have also added it.

Possibly the 2 changes require a description that I have not included and that I am not able to write so I ask for your help to include it. Also it would be necessary to add these 2 variables in the main Readme.


nvuillam:
Solved, will be in the next patch :) ( but already available on IDEs as they are targeting main branch ^^ )

Issue #870, opened 2021-10-18
Consider Using asdf


Kurt-von-Laven:
**Is your feature request related to a problem? Please describe.**
Mega-Linter presently maintains a wide variety of tool installation snippets, but this work is already done by [asdf](https://asdf-vm.com/guide/getting-started.html).

**Describe the solution you'd like**
Use asdf in tandem with .tool-versions files to install all asdf-supported tools. Write asdf plugins for any tools we need to install that don't support asdf.

**Describe alternatives you've considered**
Leave the code base as is.

**Additional context**
This change should not affect users of Mega-Linter, only contributors.

Kurt-von-Laven:
We've been using asdf for a while now across different operating systems and toolchains, and I have grown ever more sold on the idea its use could greatly simplify MegaLinter. It can serve as the base of the pyramid, with any of its enormous array of plugins layered on top (most notably almost every major package manager), and then each plugin can install the appropriate versions of the appropriate dependencies. It is a very flexible and extensible approach to dependency management.

PR #1553, opened 2022-06-23
ARM64 docker images

bdovaz:
**Edit: 05/01/2022**

I have been working a bit on this PR and thinking....

I understand that sooner or later it will get to the point where all linters are amd64/arm64 compatible either because those linters start to be arm64 compatible or because development of them has been abandoned and they end up being removed from Megalinter.

But as of today as we know, this is not the case and we don't know how long it will take..... 

## Support of runners on arm64

Windows: https://github.com/actions/runner-images/issues/6175

Linux: https://github.com/actions/runner-images/issues/5631

macOS: https://github.com/github/roadmap/issues/528

As you can see, the "closest" is macOS for Q3 of 2023. For the rest there is no ETA.

## Change from docker build to docker buildx

A first step in cross-platform image generation.

Already merged in #2199 and working for amd64.

## Supported platform descriptors configuration

With the help of @echoix is added on this PR and is that for each descriptor now allows to configure the supported platforms:

```yaml
supported_platforms:
  platform:
    - linux/amd64
    - linux/arm64
```

In some linters it is set to more platforms besides amd64 and arm64 because linter itself supports it but the reality is that initially at least we should focus on arm64 only.

This opens the door to:
1. Being able to document which descriptors work on which platforms.
2. Create a Dockerfile (or several) based on those values.

Point 2 leads us to...

## Decide how to generate the Dockerfile(s).

There are 2 options that at least I know of:

* Enter conditions in the RUN command.
  * To do this we would have to do something like this: https://stackoverflow.com/a/60518444/699056. It means that whoever makes a change in a descriptor that supports the 2 platforms has to take it into account when making modifications on the dockerfile section.
  * There is the problem that there are dependencies that are pip,apk,.... In those it is not possible to introduce a logic as with the RUN method and it would have to be solved in another way (I guess in the build.py script?).
* Generate a Dockerfile (amd64) and another one Dockerfile.aarch64 (arm64)
  *  The **build_dockerfile** function of the [build.py](https://github.com/oxsecurity/megalinter/blob/v6.17.0/.automation/build.py#L252) script should be refactored.
  * It loses sense the PR of docker buildx that is already merge since being different Dockerfile files, the build would be done in separate commands.

## Additional blocks

~~Update to .NET 6 / 7 #1680~~

## Additional blocks (external)

* misspell: I couldn't get misspell to work as it doesn't target arm (apart from the fact that the repo seems abandoned, last commit march 2018): https://github.com/client9/misspell
* Powershell: https://github.com/PowerShell/PowerShell-Docker/issues/520#issuecomment-1121468388
* rakudo-pkg: https://github.com/nxadm/rakudo-pkg/issues/96
* Chromium doesn't work in arm so I had to set PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true

There may be more...

## Testing

I have executed it in the following way (from repo root with Raspberry Pi 4 with Raspberry Pi OS 64 bits):

```bash
docker buildx build --platform linux/arm64 -f flavors/ci_light/Dockerfile .
docker buildx build --platform linux/arm64 -f flavors/cupcake/Dockerfile .
docker buildx build --platform linux/arm64 -f flavors/documentation/Dockerfile .
docker buildx build --platform linux/arm64 -f flavors/dotnet/Dockerfile .
docker buildx build --platform linux/arm64 -f flavors/go/Dockerfile .
docker buildx build --platform linux/arm64 -f flavors/java/Dockerfile .
docker buildx build --platform linux/arm64 -f flavors/javascript/Dockerfile .
docker buildx build --platform linux/arm64 -f flavors/php/Dockerfile .
docker buildx build --platform linux/arm64 -f flavors/python/Dockerfile .
docker buildx build --platform linux/arm64 -f flavors/ruby/Dockerfile .
docker buildx build --platform linux/arm64 -f flavors/rust/Dockerfile .
docker buildx build --platform linux/arm64 -f flavors/salesforce/Dockerfile .
docker buildx build --platform linux/arm64 -f flavors/security/Dockerfile .
docker buildx build --platform linux/arm64 -f flavors/swift/Dockerfile .
docker buildx build --platform linux/arm64 -f flavors/terraform/Dockerfile .
```


echoix:
Thanks for the first draft!
What I understood from working with this project the last time is that the Dockerfiles are generated by script, so if there are changes to make in the Dockerfiles, we must change the `.megalinter-descriptor.yml` like `megalinter/descriptors/kubernetes.megalinter-descriptor.yml`. Then run the `bash build.sh` script and see if the Dockerfiles produced match what is expected. 

Since you have found the changes needed, a part of the job is done!

With my secondary testing a couple months ago, that didn't finish nor end up with a pull request (after I passed around my raspberry pi 4 wasn't running in a 64 bit OS), I learned that it was possible with the buildx to cross build on my windows amd64 computer, which was quite faster with more cores and way more memory to build simultaneously the aarch64 and amd64. I didn't manage to compile up to the end though, and hadn't had a chance to look at the next problems (solving a problem one layer at the time).

The problem that I didn't have a good idea to fix, that remains, is how to change the .megalinter-descriptor.yml schema to account for the multitude of different conventions of getting and including tools for many architectures. Some might need to be in another format maybe (installed in amd64 vs Docker in aarch64). Sometimes it was a different url scheme, and not everyone names the arm 64 bit architecture the same way. Maybe one of the new active collaborators have some ideas architecture-wise?

nvuillam:
Thanks for this POC !

That looks nice, but as Dockerfiles are not written but generated (from `build.py`) , such enhancement must be done with an evolution of the architecture.

I think that having a Dockerfile_arm at the root and Dockerfile_arm files in flavors could be a good way to implement that.
And to generate such ARM-specific dockerfiles, I think that adding a new property install_arm in YML descriptor would be the way, and when generating dockerfile:
- If in the linter_descriptor there is an `install_arm` property, use it
- else use install `property`

echoix:
For
> * **megalinter/descriptors/***: add the "install_arm" field where it differs from "normal" and add whatever changes.
and
> * And in the "normal" case ignore the "install_arm" field but in the case of ARM we will have to search if the "install_arm" field exists and if not use the default "install" field.

I think we need to account that not every tool can or could be supported on first try, so if we could have a way to "check" (like in a checkbox) which tools to include in this arm flavor. This way, a compatibility table could be created showing which tools are ported.

nvuillam:
@echoix we could add a property `arm-compliant` on linter descriptors :)

echoix:
Shouldn't we be not short-sighted and inverse the logic a bit, to say we comply with these platforms, instead of considering a dichotomy amd64/aarch64(arm64)? Especially in docker architectures, I'm pretty sure some tools (flavors) like python-only could use armv7l, which is still the default Raspbian OS (most Raspberry Pis can run in 64 bit mode, but when there was no >4Gb option, there wasn't any advantage to use 64bit).

echoix:
@bdovaz I could be there for help with this PR if you need :)

bdovaz:
@echoix I'll appreciate that.

Specifically:
* Adjust the descriptor schema to add the 2 fields: arm_compliant and install_ar. I have no experience creating schemas.
* Make the settings of the "build.py" file because I have tried but I don't understand well how the Dockerfiles generation system works and which part is automatic and which is manual.
  * I have disabled validation temporarily until the json schema changes are done.
  * I have tried to create the Dockerfile_arm file of each flavor "in theory" so that it is generated as the original but I have not managed to generate more than the header.

echoix:
For sure @bdovaz, I'm looking into that this afternoon :)

Kurt-von-Laven:
I agree with @echoix that it may be better long-term to have a `supported_architectures` (or `arches`) list. I wonder how many of the linters we use have an asdf plugin with multi-architecture support. I know many of the tools we use have an asdf plugin (or, more often, a package manager with an asdf plugin).

echoix:
Could I have a little recap on asdf, or how it is used or planned to be used in Megalinter? I receive most of the issues of this repo by email, but I still don't understand it in relation to this project.

Kurt-von-Laven:
buildx does seem like a good way to go with the caveat that I always optimistically hope there is some approach that involves less work ha ha ha. Is [this](https://wiki.debian.org/Multiarch/HOWTO) the same multiarch you are referring to? #870 contains some of my thoughts on asdf's relevance to MegaLinter. [asdf](https://asdf-vm.com/) is capable of installing [all of these tools](https://github.com/asdf-vm/asdf-plugins#plugin-list). It is basically a version manager manager, but it can manage versions of most anything. asdf has expressed interest in supporting multiple architectures in the core if necessary (c.f., asdf-vm/asdf#62), but I don't know which of the plugins presently do. One of the challenges MegaLinter inherently faces is managing many dependencies from many different package ecosystems plus some with bespoke installation methods. It will simplify maintenance if we can eventually install everything that has an asdf plugin (or, better yet, a package manager that has an asdf plugin) through asdf, but the first step would be to simply install one package manager this way. I suppose one way of marrying the ideas would be to use your buildx approach to improve or create asdf plugins since then MegaLinter could leverage that uniformity to install everything with as much multi-architecture support as the plugin (and underlying tool) allows. This might also be a good way to get more hands on deck since the asdf community is presently much larger than MegaLinter's, and I assume many people who don't use MegaLinter share your woes getting things working on various architectures. If there is some tool other than asdf that is already better suited to your goal, I imagine someone there will educate us, particularly since it brings together a diverse cross-section of people using a huge variety of toolchains.

echoix:
> Is this the same multiarch you are referring to?

Nope, the multi arch in Docker, where there an image's manifest contains different images for many cpu architectures. https://docs.docker.com/desktop/multi-arch/

bdovaz:
> For sure @bdovaz, I'm looking into that this afternoon :)

Have you finally started working on it? 

Likewise you would have to clearly define how you want to make the changes.

As far as I have understood it would be to put a "supported_architectures" to be able to distinguish the installation commands and that for the future it is better designed.

My doubt is at what level would it be? Below "install", above "install"?  I also have doubts with the name "amd64" and "arm64", opinions?

Example 1:

```yaml
linters:
  - linter_name: ansible-lint
    install:
      supported_architectures:
        amd64:
          pip:
            - ansible-lint==6.0.2
        arm64:
          pip:
            - ansible-lint==6.0.2
```

Example 2:

```yaml
linters:
  - linter_name: ansible-lint
    supported_architectures:
      install:
          amd64:
            pip:
              - ansible-lint==6.0.2
          arm64:
            pip:
              - ansible-lint==6.0.2
```

echoix:
(Sorry, I started this message on Monday but never had the opportunity to finish it)

For the name of the architecture used, it's sure that we will be using the names used by Docker, it greatly simplifies writing multi-architecture Dockerfiles.
Ideally, the same Dockerfile can be used for multiple architectures. That way, `buildx build --target` can be used directly for all architectures (like many multi-arch image https://docs.docker.com/desktop/multi-arch/)

I drafted out scenarios (without committing anything yet), but couldn't find a way that would be a breaking change. The most common problem is when there is a URL that doesn't use the same architecture names.

One of the scenarios that wouldn't be as breaking, would be that we keep the `install` property with their pip, docker, apk, etc children fields, but add some architecture-specific overrides.
So in this case, `supported_architectures`, could have the architectures, and any overrides in the same format as `install`.
One simple override I think could be used to make architecture-independent Dockerfiles is having a mapping of the architecture name, for mapping the shellcheck architecture URL that uses aarch64. Otherwise, we could always use inline shell `case`/`esac` construct, in the original `RUN` command.

A scenario that would be breaking, is explicitly having all instructions for each platform, but doesn't help reuse layer caching.

In the same line of thought, common multi-arch Dockerfile patterns use multi-stage Dockerfiles to be able to execute some build in the builder with one architecture, and for the final layer(s), have them in the target instruction set, so this speeds up building.

echoix:
I completed thinking out the JSON schema, and this should cover everything. I opened a PR to your branch, so it can me merged into this PR.

I started one converrsion of megalinter descriptor, and another day I would need to work with the build.py. We shouldn't need to have multiple files if using buildx and proper multi-stage builds, but in the meantime, with some tricks like the following, we could not even need any overrides, and just adapt the descriptor's install commands.

https://github.com/BretFisher/multi-platform-docker-build/blob/cdd1bb1135554c41f5a4b0f09c8e912eae6a5022/example-tini/Dockerfile#L11-L20

https://github.com/crazy-max/docker-docker/blob/2d6870eecee16f15b7f3ef8140ed975d38cb2866/Dockerfile-20.10#L26-L37

(from https://github.com/BretFisher/multi-platform-docker-build)


As for the changes required in build.py, apart from the logic of building or not the flavors for each platform, to be able to efficiently use caching and parallelizing parts of builds that dont change, the idea would be to have small stages, perhaps having empty stages for tools unavailable on a platform where we could split the fetches and unpack, and/or merge back with two possible paths, one per target (the part where there is a series of copies). If the concept of little independent stages could be implemented, this would allow to have the cache still valid even if one tool has a version upgraded, and only in the copy stages at the end that the cache would be invalidated. See the idea below from this post https://itnext.io/building-multi-cpu-architecture-docker-images-for-arm-and-x86-1-the-basics-2fa97869a99b
> ### Multi-Arch Dockerfile with Multi-stage Build
> You might not always find the base image that supports all target platforms that you plan to build. With [the multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/), it’s possible to conditionally select different base images depending on the current building platform (by checking TARGETOS & TARGETARCH ARG) in your Dockerfile.
> 
> Here is a complete Dockefile example:
> 
> ```Dockerfile
> # Declare stage using linux/amd64 base image
> FROM --platform=linux/amd64 docker.elastic.co/elasticsearch/elasticsearch:6.8.22 as stage-amd64
> # Commands run only for linux/amd64 target platform
> RUN yum -y install sudo zip
> RUN find modules -type d -name "x-pack-*" -exec rm -r {} +
> COPY --chown=elasticsearch:elasticsearch component/elasticsearch.yml /usr/share/elasticsearch/config/
> # Declare stage using linux/arm64 base image
> FROM --platform=linux/arm64 data61/elasticsearch:6.8.22 as stage-arm64
> # Commands run only for linux/arm64 target platform
> RUN apt-get update && apt-get install -y --no-install-recommends sudo zip && rm -rf /var/lib/apt/lists/*
> COPY --chown=elasticsearch:elasticsearch component/elasticsearch-arm64.yml /usr/share/elasticsearch/config/elasticsearch.yml 
> # Declare TARGETARCH to make it available
> ARG TARGETARCH
> # Select final stage based on TARGETARCH ARG
> FROM stage-${TARGETARCH} as final
> # Commands run at final stage for all target platforms
> ADD component/setup.sh /setup.shCOPY --chown=elasticsearch:elasticsearch component/wn_s.pl /usr/share/elasticsearch/config/analysis/
> COPY --chown=elasticsearch:elasticsearch component/regionSynonyms.txt /usr/share/elasticsearch/config/analysis/
> CMD ["/setup.sh"]
> ```

This reading kinda has diagrams of what I'm trying to explain too https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

The changes needed for this is simply handling the way the array of `dockerfile` commands in the descriptor are split, so that multiple froms (with all their commands) could stay together instead of checking "if starts with FROM, bump the complete line at the dockerfile's header, else if check if the line starts with arg, else treat it normally".

bdovaz:
@echoix I have merged your PR and updated my changes.

I am encountering a problem and that is that there are in several cases where the changes that need to be made is not at the linter level but at the descriptor level.

Look at this change:

https://github.com/oxsecurity/megalinter/pull/1553/files#diff-614f3cce8a8bb565c2d879cc11efa6e21100fb2d32ba39ad2ab25aa446a67be8

Right now it obviously does not comply with the schema because the changes you made only apply to the level of a linter, but somehow we have to accommodate the whole descriptor.

Can you help? Thanks.

echoix:
Oh, I see now how it is used there, it wasn't clear why there was a semi-duplicate. With this, I understand how it is used and why we would need it there too. 

I'll be able to update it this way. Leave me a bit of time though, I'm not at my computer :)

bdovaz:
I have also noticed that in many cases it is simply changing the architecture in the file name of the binary and it is quite ugly to have to duplicate all the commands just for that minimal change.

If somehow there was a way to pass arguments to the install commands like:

```yaml
install:
  dockerfile:
    command:
      - ARG PWSH_VERSION='latest'
      - ARG PWSH_DIRECTORY='/opt/microsoft/powershell'
      - |
        RUN mkdir -p ${PWSH_DIRECTORY} \
            && curl --retry 5 --retry-delay 5 -s https://api.github.com/repos/powershell/powershell/releases/${PWSH_VERSION} \
                | grep browser_download_url \
                | grep linux-{{arch}} \
                | cut -d '"' -f 4 \
                | xargs -n 1 wget -O - \
                | tar -xzC ${PWSH_DIRECTORY} \
            && ln -sf ${PWSH_DIRECTORY}/pwsh /usr/bin/pwsh
    args:
      - arch=arm64
supported_platforms:
  platform:
    - linux/amd64
    - linux/arm64
  install_override:
    - platform: linux/amd64
      install:
        dockerfile:
          args:
            - arch=alpine-x64
    - platform: linux/arm64
      install:
        dockerfile:
          args:
            - arch=arm64
```

Where those arguments are replaced by the templating engine used (if one is used) when copying these descriptor and linter fragments to the resulting Dockerfile.

bdovaz:
@echoix I have merged your two PRs, thanks!

Pending tasks:
- [ ] Create a CASE in "install > docker" for the cases where you only vary the architecture string in the binary (case of at least bash and powershell at the moment).
- [ ] Create the necessary code to read the new properties created in the descriptors and linters to add to the resulting Dockerfile. They are not yet being used.
- [ ] With buildx generate each image in each architecture and be sure that build is done correctly.
- [ ] Modify the script that generates the flavors/linters documentation to indicate per architecture what works and what does not.

I also don't know how you tested the bash one as you mentioned, plus I see you listed more platforms than amd64, arm64 and arm/v7.

echoix:
> I also don't know how you tested the bash one as you mentioned, 
That one (the bash `case` in the RUN instruction) I didn't formally test, since it will be possible to test more thoroughly when we will be doing the build tests. There is no apparent reason to not work, it's theoretically correct.


> plus I see you listed more platforms than amd64, arm64 and arm/v7.
I consider you're talking about bash-exec, where I listed more architectures. Here I listed all the architectures that the image supports, since it's native bash in this linter's case. My reasoning was to include all the supported architectures that the linters support in the descriptors support, so that a user with who would want a different platform could build it. We could (or not) create the Dockerfile(s) for every architecture each time, even if we only build the containers for the platforms and flavours that are common enough. (I suspect that the Python flavour could be such a candidate, so we could leave the door open). Having more supported_platforms doesn't mean that we will build containers for each.
Since to know the installation steps we go open each tool's website/repo and look at their releases, we know by the same time what platforms it supports.

SoumayaMauthoorMOJ:
I don't have the skills for making changes but as I told @echoix happy to do some testing on my M1 once ready! I've been using the oxsecurity/megalinter-python:v6 docker image and linting+formatting works as expected apart from the speed (I estimate about 5 times slower than on Mac Intel)

bdovaz:
As I said, I need support in this PR to unlock it because it is quite a lot of work and I am not able to tackle it alone.

Kurt-von-Laven:
Is there a way this PR can be decomposed into more bite-sized pieces that don't need to all be landed atomically? My experience in the open-source world has been that single large atomic units of work often tend to be indefinitely deprioritized.

echoix:
I've been working on the build.py this weekend. I thought of an element of this PR (that is still left to be done) that could be part of a separate PR and be used before this is ready that could reduce the risk of changing too much. 

If the project could already switch to use BuildKit (the Docker backend) either with a env variable or changing the places where `docker build` is called to `docker buildx` (preferable, as we can use the `--platform` argument), it would help, as it will be necessary to cross-build the containers for other architectures. Like if it could be running with it for a week or two only for the amd64 images it would make it easier to add another platform to it.

nvuillam:
The docker build part is inherited from super-linter, good old ugly bash, so if you want to make propositions to enhance it, it will be greatly appreciated :)

bdovaz:
Does anyone want to work on this PR? As I already said, I don't see myself able to face all this development in a technology that I don't master.

SoumayaMauthoorMOJ:
Is there perhaps an interim solution based on creating an ARM64-compatible flavor? So I can install megalinter on my Mac M1 and use the standard python flavor, it's just extremely slow (>5mins). However my colleague @oliver-critchfield did some investigation and cut it down to <90s with the following config:

```
ENABLE:
  - COPYPASTE
  - DOCKERFILE
  - GHERKIN
  - JSON
  - MARKDOWN
  - PYTHON
  - REPOSITORY
  - SQL
  - SPELL
  - YAML
DISABLE_LINTERS:
  - MARKDOWN_MARKDOWN_LINK_CHECK # Disabled as links requiring an account break this
  - PYTHON_MYPY
  - PYTHON_BANDIT
  - PYTHON_PYRIGHT
  - REPOSITORY_CHECKOV # This scans cloud infra cofigurations, so not applicable here
  - REPOSITORY_DEVSKIM
  - REPOSITORY_SEMGREP
  - REPOSITORY_SYFT
  - REPOSITORY_TRIVY
  # DISABLE LINTERS WHICH ARE SLOW-RUNNING ON M1 mac
  - COPYPASTE_JSCPD
  - REPOSITORY_SECRETLINT
  # DISABLE DOCKERFILE AS M1 COMPATIBILITY ISSUES
  - DOCKERFILE_HADOLINT
 ```

**Comment by bdovaz** [comment]
No stale.

What do we do with this one? As I said, it's too big for me (I have no experience in Python). As @Kurt-von-Laven says, maybe we should rethink this PR in other smaller PRs to unlock it...

**Comment by bdovaz** [comment]
@Kurt-von-Laven @echoix @nvuillam I have updated the first post with the current status and to create discussion with the topic of generating the Dockerfile(s).

I would appreciate your input as this is something that cannot be moved forward without other points of view.

**Comment by nvuillam** [comment]
@bdovaz from the examples I saw in articles on the net, it's simpler to have one dockerfile for all archs, and the this single dockerfile is used by the 2 builds that buildx makes in parallel

So in my opinion, all archs should be managed in a single `install` property in descriptors.

The supported platforms is good, it will allow at runtime to notify users that for some linters, arm64 is not supported for example (this would require some check in Linters.py probably)

**Comment by bdovaz** [comment]
@nvuillam I agree that it is easier and more logical with a single Dockerfile. But what do you tell me about what I put in the first post?

> There is the problem that there are dependencies that are pip, apk,.... In those it is not possible to introduce a logic as with the RUN method and it would have to be solved in another way (I guess in the build.py script?).

Do you see it feasible?

**Comment by nvuillam** [comment]
@bdovaz what could be wrong with apk npm pip etc ? their command line to call is the same for amd64 / arm64 , no ? ^^

**Comment by bdovaz** [comment]
> @bdovaz what could be wrong with apk npm pip etc ? their command line to call is the same for amd64 / arm64 , no ? ^^ I assume `pip install` , `npm install` etc... manage themselves to detect where they are and provide the appropriate sources / binaries ?

@nvuillam what I mean is that with `install > dockerfile` you can put conditions (to filter by arch) as we have already mentioned but with `install > npm/pip/apk/...` not. If I am not wrong, in that case the descriptor only allows you to add the package name and nothing else. That's why without touching the `build.py` script to alter the generation of the Dockerfile files, I don't see a platform-dependent way to add or remove package entries from npm/pip/apk/....

**Comment by bdovaz** [comment]
@nvuillam I'm making progress on getting conditions into dockerfile RUN commands and I'm running into strange discrepancies in terms of command tabs where being conditions, it's important.

Example descriptor with condition:

image

Result:

image

If you notice, in the descriptor I had to add in the content inside the condition 5 spaces so that later in the Dockerfile file it translates it to 2? I don't understand why I had to put more spaces...

**Comment by nvuillam** [comment]
> > @bdovaz what could be wrong with apk npm pip etc ? their command line to call is the same for amd64 / arm64 , no ? ^^ I assume `pip install` , `npm install` etc... manage themselves to detect where they are and provide the appropriate sources / binaries ?
> 
> @nvuillam what I mean is that with `install > dockerfile` you can put conditions (to filter by arch) as we have already mentioned but with `install > npm/pip/apk/...` not. If I am not wrong, in that case the descriptor only allows you to add the package name and nothing else. That's why without touching the `build.py` script to alter the generation of the Dockerfile files, I don't see a platform-dependent way to add or remove package entries from npm/pip/apk/....

My hope is that none of those packages (apk, npm, pip, gem)  will need conditions, let's try that first and see if it works ^^

**Comment by nvuillam** [comment]
> @nvuillam I'm making progress on getting conditions into dockerfile RUN commands and I'm running into strange discrepancies in terms of command tabs where being conditions, it's important.
> 
> Example descriptor with condition:
> 
> image
> 
> Result:
> 
> image
> 
> If you notice, in the descriptor I had to add in the content inside the condition 5 spaces so that later in the Dockerfile file it translates it to 2? I don't understand why I had to put more spaces...
> 
> The logic regarding the discrepancy I understand is this but I also don't see very clear why (I don't have much experience with python):
> 
> https://github.com/oxsecurity/megalinter/blob/main/.automation/build.py#L312

build.py just appends run instructions to decrease the number of dockerfile step, i fon't know what could change indentation (maybe something in your IDE ?)

Bit it works anyway, no ? ^^

**Comment by nvuillam** [comment]
Note: don't bother with windows arch, docker for windows can run linux images so no need ^^

**Comment by echoix** [comment]
I'm writing a response, you're faster than me..!

**Comment by nvuillam** [comment]
_Ignore them and only indicate in the descriptor: amd64 / arm64._

Plz keep the exact identifiers that can be sent to docker buildx ^^

**Comment by echoix** [comment]
I understand what @bdovaz means. Last time I looked for what linters were available, sometimes for a certain arch we needed a different installation method. So overrides are necessary, and we can't simply use the same install when it comes to grouping the packages.
I solved this problem conceptually this fall, last time I really took the time to sit on this. But my code was not ready to commit, and now, it's way out of sync with the main branch.

I will still commit and push my branch so you can take a look at it.

Yes, everything is played in the build script.

My schema update, if the latest one was ever included in this PR, had a way to add overrides by supported platforms. Where I wasn't finished, and found that something usually easy for me (python) became stuck, was to properly handle the cases of the default "properties" of a linter. I didn't finish making sure that the dictionary/object falled back to what was the expected behaviour (I was still getting some nulls and some deep-copy/shallow-copies errors (work for the first but not the rest, and had a tough time remembering correctly dict-comprehensions).

Then to integrate the real build.py > generates Dockerfiles, I see now two ways (instead of just one). Since we are running with buildx, no need to really care about making a plain Dockerfile work.

1. Generate almost full Dockerfiles for each arch inside the same, and use the names in the multi-stage to select which "branch" to keep in the final build for that platform. Since now the approach is to shove everything in less layers, my previous "wish" of user multi-stage builds efficiently doesn't seem to fit with that view.

```mermaid
graph TD
    A["Froms: #FROM__START ... #FROM__END"] -->|Base image - generic| B("FROM python:3.10.4-alpine3.16") 
    B --> |"(rename to)"| BB["FROM python:3.10.4-alpine3.16 as megalinter-common"]
    BB --> C[Args: #ARG__START...#ARG__END]
    C -->D[Other generic stuff]
    D -->|"For targetplatform linux/amd64 :: no variant"| E[FROM megalinter-common as build-linux-amd64--ml]
    D -->|"For targetplatform linux/arm64 :: variant is empty even if /v8 is added"| F[FROM megalinter-common as build-linux-arm64--ml]
    D -->|"For targetplatform linux/arm/v7 or linux/arm :: (32 bit)"| G[FROM megalinter-common as build-linux-arm-v7-ml]
    E --> H["FROM build-${TARGETOS}-${TARGETARCH}-${TARGETOS}-ml as megalinter-arch-specific"]
    F --> H
    G --> H
    H --> |"Remaining COPY, ARGS, ENV, LABELS, etc."| I["End of Dockerfile"]
```
  
2. Hardcode some logic inside the commands to run when installing packages to evaluate the target platform when combining the install commands. Or even more drastic, but simple: Each target platform has all the packages repeated for every platform, but there is a shell `if` that executes the correct list. Same thing for Dockerfiles. 


See this where there are outputs for the TARGETPLATFORM, TARGETARCH, TARGETVARIANT : https://github.com/BretFisher/multi-platform-docker-build
See this for the conditional in Dockerfile: https://www.docker.com/blog/advanced-dockerfiles-faster-builds-and-smaller-images-using-buildkit-and-multistage-builds/#4374

Maybe just a reminder that (maybe it won't be a problem if the variables are globally defined when using buildx, but for args it still applies), if we want to have arguments inside another stage in a multi-stage build, we need to "redeclare" them in each stage (under the FROM), since they are scoped. 
https://docs.docker.com/engine/reference/builder/#scope
https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

And remember, "[BuildKit](https://docs.docker.com/build/buildkit/) only builds the stages that the target stage depends on." -> https://docs.docker.com/build/building/multi-stage/#differences-between-legacy-builder-and-buildkit

nvuillam:
Interesting thoughts :)

I didn't know that we could use if / else outside of RUN instructions:)

Anyway... i'd really like to have a single dockerfile called woth buildx so on docker hub we'll have a single image containing 2 arch layers 

If for some reason, some apk / pip / npm / gem packages require conditioning, it is still possible to not declare them as package, but to define their installation within a RUN instruction (containing those conditions)

echoix:
> I didn't know that we could use if / else outside of RUN instructions:)

Yes, a bit, but everything must exist at the start of the build, and we can't have variables after the "as .....". Buildx will try to run the least possible and will prune what isn't needed for the job, so it acts like a if/else

> Anyway... i'd really like to have a single dockerfile called woth buildx so on docker hub we'll have a single image containing 2 arch layers 

Yep, all my ideas still have one Dockerfile and will publish as the same name (multi platform)

> If for some reason, some apk / pip / npm / gem packages require conditioning, it is still possible to not declare them as package, but to define their installation within a RUN instruction (containing those conditions)

That was like my option 2, where there is some bash logic in a run command, so it remains linear as now.
Branches/layers when we use buildx is really interesting, since if there isn't any dependencies, they'll be done in parallel. GitHub hosted action runners have 2 cores, so it's maybe not optimal, but at least

nvuillam:
I already thought about multistage build then copy , we could do at least one stage by package manager and copy node_deps, venvs, etc...

That would increase build performance, but i think such idea can be do after this PR, to avoid having dozens of conflicts with it :)

echoix:
Yep. I was thinking the same for the series of copy at the end, to branch them together and batch copy one intermediate (it will then only have one layer) instead of like 12). I'm just not sure if there is any overwriting when we use COPY

echoix:
@bdovaz I forgot to add the link of the partial state of my branch: https://github.com/echoix/megalinter/tree/multi-arch-build-dockerfile

bdovaz:
@echoix I am from a Windows machine on local trying to run the main Dockerfile:

`docker buildx build --platform linux/arm64 -f Dockerfile .`

I'm progressing little by little with the conditions that I'm adding by architecture in the descriptors in the dockerfile section.

But at a certain point in the execution I get this error:

image

Do you know if there is a solution? I just don't know if qemu emulation is reliable or there are in certain cases like this that it is not possible to solve the problem.

For this PR one of the problems that we will have that I put in the first post is that we do not have a native runner in ARM so I do not know how we will be able to test this PR to merge it ....

bdovaz:
@nvuillam @echoix for the runners issue I see that there is _Cirrus CI_ that apparently allows to add runners to github and as I understand "I think" it is free for projects like Megalinter:

https://cirrus-ci.org/blog/2022/11/03/github-actions-on-m1-via-cirrus-runners/

https://cirrus-ci.org/pricing/

The Rust repository takes this into account:

https://github.com/rust-lang/rust/issues/73908

echoix:
@bdovaz could I see a little more context on that missing lib?

echoix:
Buildx with qemu is what I'm expected will be used on GitHub, there's no problem there. It will be slower than native, but native isn't available easily. It's a complete other structure to build multi-arch Docker images on different computers, let alone on different CI. 

So it's not really a problem. Of course I'd prefer validating at least once a native build for the arm64 image, but we have access to way faster amd64 machines

nvuillam:
> @nvuillam @echoix for the runners issue I see that there is _Cirrus CI_ that apparently allows to add runners to github and as I understand "I think" it is free for projects like Megalinter:
> 
> https://cirrus-ci.org/blog/2022/11/03/github-actions-on-m1-via-cirrus-runners/
> 
> https://cirrus-ci.org/pricing/
> 
> The Rust repository takes this into account:
> 
> https://github.com/rust-lang/rust/issues/73908

What is the runner issue ?

nvuillam:
@bdovaz in order to avoid windows problems with docker build using arm64, maybe just use GitHub CI to do that ?

echoix:
> What is the runner issue ?

I don't know either, but if it's the screenshot of the single line of a build log, my answer was appropriate

echoix:
> @bdovaz in order to avoid windows problems with docker build using arm64, maybe just use GitHub CI to do that ?

A local docker build would probably still be faster, it isn't some really powerful runners on GitHub. Especially if he manages to use some little amount of Docker cache between builds. I'm not sure with the packed layers (all in one big command) how much cache that can be used is left though.

bdovaz:
I have encountered another blocking problem: https://github.com/docker/buildx/issues/59.

nvuillam:
@bdovaz  Seems there is a workaround -> https://github.com/docker/buildx/issues/59#issuecomment-1168619521

nvuillam:
You really should merge this branch into a branch in the main repository, because you have more jobs ( docker images for individual linters) so it will be much faster for testing with CI :)

echoix:
> You really should merge this branch into a branch in the main repository, because you have more jobs ( docker images for individual linters) so it will be much faster for testing with CI :)

Since I'm collaborator, I can create a branch from main in the oxsecurity/megalinter repo, and change the target branch of this PR to that new branch. We could then merge the PR, and have a branch inside oxsecurity/megalinter to work on, that at the end we would need to make a new PR to merge into main.

To continue editing inside a branch of oxsecurity/megalinter, @bdovaz would need to be able to push to that branch. If not, he would need to create a new PR with the exact same problem.

And lastly, where would the conversation about the work go? We could immediately open a PR from the branch inside megalinter, so at least we could see the CI results as a PR.

nvuillam:
@echoix > @bdovaz is also a collaborator so he can do that too ^^

echoix:
Ok great! so I think it might be a good idea!

bdovaz:
> @bdovaz it seems there is a workaround -> [docker/buildx#59 (comment)](https://github.com/docker/buildx/issues/59#issuecomment-1168619521)

@echoix can you help me to fix it?

There are several lines here where we should make that change but I don't have much knowledge of bash and I don't know how to change it so that it doesn't affect the following:

In upload-docker.sh:

https://github.com/bdovaz/megalinter/blob/feature/arm-docker-image/.automation/upload-docker.sh#L319

image

In several workflows like this:

https://github.com/bdovaz/megalinter/blob/feature/arm-docker-image/.github/workflows/auto-update-linters.yml#L61

image

echoix:
> There are several lines here where we should make that change but I don't have much knowledge of bash and I don't know how to change it so that it doesn't affect the following:
> 
> 
> 
> In upload-docker.sh:
> 
> 
> 
> https://github.com/bdovaz/megalinter/blob/feature/arm-docker-image/.automation/upload-docker.sh#L319
> 
> image
> 


The `2>&1` is a output redirection that sends the output to somewhere. Program output can be stdout and stderr. It's possible to redirect and output to a file. 2 is a file descriptor for stderr. 1 is a file descriptor for stdout. It's also possible to redirect an output to another stream, instead of a file. And that's what is here. Stderr is redirected to stdout. https://stackoverflow.com/a/818265

I was thinking about this, one day we could let go of bash specific syntax. Is there still a use case where the building of the images of megalinter isn't done on GitHub actions? If we aren't able to understand it well, well, we aren't maintaining it. 

Maybe a quick search to see if the github action logs already include both outputs. 
However, the important shell syntax is what follows, https://github.com/bdovaz/megalinter/blob/ea67c7777ed2e0e4e7f8642b0c6d29e0c9936017/.automation/upload-docker.sh#L324

The `$?` is a special variable that is equal to the exit status of the last executed command. I don't really like relying on bashisms. If someone doesn't know about it, it's really hard to follow and debug. 
 https://stackoverflow.com/a/7248048


> 
> In several workflows like this:
> 
> 
> 
> https://github.com/bdovaz/megalinter/blob/feature/arm-docker-image/.github/workflows/auto-update-linters.yml#L61
> 
> 
> 
> image


Here I'm not sure what construct you want to know, is it the pipe character (`|`)?

Here, that line calls the command normally, and at the end, the pipe sends the stream to something else as an argument, (the read line thing), which is a loop that takes each line outputted ($line being an argument), and instead runs `echo`, to print out a message to the console, and that message is the current date and time with that format, a pipe character (it's inside a string, it's not the operator), and adds the contents of the variable $line. In short, it prepends a time stamp to each line output. Since GitHub action logs can show the time stamps (maybe it wasn't possible before), the rest of the line after `load .` can either be kept, or simply removed without impact.

bdovaz:
@echoix are you clear about the change to make without breaking the commands below in the 2 cases I have put? In case you want to make help me with a PR in my fork or if you can tell me what to change it to exactly.

Thanks!

From this change could already be done the merge issue that you were talking about in which as I said, I prefer that you give me a hand because of my limited knowledge in git.

echoix:
For this line: https://github.com/bdovaz/megalinter/blob/ea67c7777ed2e0e4e7f8642b0c6d29e0c9936017/.github/workflows/auto-update-linters.yml#L61
It's non breaking to change to 
```yaml

        run: docker buildx build --platform linux/amd64,linux/arm64 --build-arg "BUILD_DATE=${BUILD_DATE}" --build-arg "BUILD_REVISION=auto_update_${GITHUB_SHA}" --build-arg "BUILD_VERSION=auto_update_${GITHUB_SHA}" --no-cache -t oxsecurity/megalinter:auto_update_${GITHUB_SHA} --load .

```

That is, removing `| while read line ; do echo "$(date +'%H:%M:%S')| $line"; done; `

The other one, I didn't really specify a change. It's answering what it is. I don't like it, but it might be deep for a first shot.

echoix:
@bdovaz 
> From this change could already be done the merge issue that you were talking about in which as I said, I prefer that you give me a hand because of my limited knowledge in git.

Oh I think I understand. You mean the suggestion to target a branch inside the oxsecurity/megalinter repo, and continue working on it there?

I'll try to take a look this evening, it's still a workday in my timezone and gotta go back ;)

nvuillam:
Something i thought about (maybe for a later PR... or this one ? ) is to get rid of our code building docker, as it seems that there is a github action from docker that does that very well using buildx, and the neighbours of Super-Linter are using it

https://github.com/github/super-linter/blob/bb170af75d406d562ae56771946053adeed41ca6/.github/workflows/cd.yml#L59

https://github.com/docker/build-push-action

nvuillam:
Another thing that they do well is that when releasing, they don't build again from scratch like we do, but they take the beta images and retag them

echoix:
Same thing for Login Docker and Setup QEMU. We can have multiple login docker for multiple registries, and the post action to logout with chain themselves. I tried it yesterday on another repo: 
https://github.com/echoix/editorconfig-checker/blob/66b890c81c54c282068e5b1591049cb68eeb75ad/.github/workflows/release.yml#L29-L45

```yaml
      - # Add support for more platforms with QEMU (optional)
        # https://github.com/docker/setup-qemu-action
        name: Set up QEMU
        uses: docker/setup-qemu-action@v2
      - name: Docker Setup Buildx
        uses: docker/setup-buildx-action@v2
      - name: Login to Docker Hub
        uses: docker/login-action@v2
        with:
          username: ${{ secrets.DOCKERHUB_USERNAME }}
          password: ${{ secrets.DOCKERHUB_TOKEN }}
      - name: Login to GitHub Container Registry
        uses: docker/login-action@v2
        with:
          registry: ghcr.io
          username: ${{ github.actor }}
          password: ${{ secrets.GITHUB_TOKEN }}
```

image

Kurt-von-Laven:
> Maybe a quick search to see if the github action logs already include both outputs.

They do, yes, but maybe the intention of redirecting stderr to stdout was intended to guarantee stable output order (c.f., isaacs/github#1981) or something else?

> I don't really like relying on bashisms. If someone doesn't know about it, it's really hard to follow and debug.

So true, and Bashisms are also unusually difficult to Google.

> Something i thought about (maybe for a later PR... or this one ? ) is to get rid of our code building docker, as it seems that there is a github action from docker that does that very well using buildx, and the neighbours of Super-Linter are using it
> https://github.com/docker/build-push-action

Good point! Yes, using build-push-action is the best practice I'm aware of here.

echoix:
@bdovaz I've created that dev branch at https://github.com/oxsecurity/megalinter/tree/dev/multi-platform-images. I let you create a new Draft pull request from that branch to continue discussions there, and to have CI there, and it won't be a fork.
We will announce in the PR the location of the new discussion once done.

PR #1561, opened 2022-06-24
fixed a bug that caused `--containername` to not work

### LINKED ISSUE #1560: When `--containername` option is used, the container isn't given the specified name
**Describe the bug**
Previously I crated a feature request (#1557) followed by a pull request (#1558) to allow naming of MegaLinter Docker container. Unfortunatlly I didn't test it well, so the feature doesn't actually work (although it doesn't break anything).

**To Reproduce**

1. Execute a command `npx mega-linter-runner@5.15.1-beta202206241616.0 --containername=TEST` on any projet
2. When done (or when it's still runnning) execute `docker ps -a`
3. Actual result: there is a container with image name `megalinter/megalinter:v5` but the container name is random (expected `TEST`).

**Expected behavior**
Expected to see a a container with image name `megalinter/megalinter:v5` with container name `TEST`.

**Additional context**
The `5.15.1-beta202206241616.0` version is the one that got cretaed when #1558 was merged to `main`. I have the fix coming, will create a PR soon.

dmitri-trofimov:
Fixes #1560 - When `--containername` option is used, the container isn't given the specified name 

## Proposed Changes

1. Fixed the order of command line arguments (the `--name` should appear before `-v`)
2. Changed the source of container name from `commandArgs.contanername` to `options.containername`.

### Author/Contributor

### Reviewing Maintainer


mega-linter-runner/lib/runner.js
```diff
@@ -113,13 +113,12 @@ ERROR: Docker engine has not been found on your system.
...
     // Build docker run options
     const lintPath = path.resolve(options.path || ".");
-    const commandArgs = [
-      "run",
-      "-v",
-      "/var/run/docker.sock:/var/run/docker.sock:rw",
-      "-v",
-      `${lintPath}:/tmp/lint:rw`,
-    ];
+    const commandArgs = [ "run" ];
```
Kurt-von-Laven, line 116: Nitpick: consider removing whitespace inside square brackets for consistency with other array literals.

mega-linter-runner/lib/runner.js
```diff
@@ -113,13 +113,12 @@ ERROR: Docker engine has not been found on your system.
...
+    if (options.containername) {
+      commandArgs.push(...["--name", options.containername]);
+    }
+    commandArgs.push(...["-v", "/var/run/docker.sock:/var/run/docker.sock:rw"]);
+    commandArgs.push(...["-v", `${lintPath}:/tmp/lint:rw`]);
```
Kurt-von-Laven, line 121: Nitpick: I like the clarity of keeping each argument on its own line; this clarity can be retained with marginally less code in a single push:

```js
commandArgs.push(…[
  "-v", "/var/run/docker.sock:/var/run/docker.sock:rw",
  "-v", `${lintPath}:/tmp/lint:rw`
]);
```

mega-linter-runner/lib/runner.js
```diff
@@ -134,9 +133,6 @@ ERROR: Docker engine has not been found on your system.
         commandArgs.push(...["-e", envVarEqualsValue]);
       }
     }
-    if (commandArgs.containername) {
-      commandArgs.push(...["--name", commandArgs.containername]);
```
Kurt-von-Laven, line 138: This is most likely a matter of my own ignorance, but I am not seeing why the previous order of options would work any better or worse than the new order from reading the [`docker run`](https://docs.docker.com/engine/reference/commandline/run/) documentation. Would you mind adding a brief explanation of why this change fixes the bug to the commit message?

Kurt-von-Laven (changes requested):
Thank you for the quick bug fix! Some minor suggestions and one question for you.

mega-linter-runner/lib/runner.js
dmitri-trofimov: To be honest, I tried to find the relevant thing in docs, but couldn't. However, if I move the `--name` argument after the `-v` it no longer works :(

Kurt-von-Laven: Sounds like the root cause is Docker’s documentation then. Thank you for educating me.

PR #1563, opened 2022-06-24
Allow removal of docker container when done

### LINKED ISSUE #1562: Add an option to remove docker container when linting is done
Running `megalinter` multpiple times on a local computer causes it to create many containers. Often we don't need to keep them and would like to remove them as soon as the linting is done.

I would like a command line argument (like `--removecontainer`) that would allow me to let `meagalinter` know that it should cleanup after itself and remove the container when done linting. `megalinter` could do it by adding `--rm` to the `docker run`.

It is possible to automate the container clenaup in a different way - list containers, find the one that was just executed and remove it. But it does seem like too much complexity, considering Docker already has the `--rm` option.

**Additional context**
I can create a PR for that shortly.

dmitri-trofimov:
Fixes #1562 

## Proposed Changes

1. Add a command-line option `--removecontainer` to the megalinter runnner
2. The runner will pass the `--rm` option to `docker run` when the `--removecontainer` is specified
3. The `--rm` option ensures the removal of docker container (but not the image) when linting is done.

### Author/Contributor

### Reviewing Maintainer


Kurt-von-Laven (changes requested):
I run `docker system prune` from time to time to deal with this, which works well for power users since it gives control over when the cleanup takes place, but forces additional maintenance onto the developer, and there is no guarantee that every MegaLinter user is familiar with Docker. I don’t have any particular comments on this PR besides the ones I already put on #1561, but my personal preference would be to ferry arguments through to Docker rather than mirror a subset of its API. Let us give @nvuillam a chance to weigh in before I send you off in a poor direction though.

Kurt-von-Laven:
It looks good to me to merge once v6 is released!

nvuillam:
@dmitri-trofimov please could you upgrade your branch ? :)

nvuillam (approved):
Good, thanks for your contribution :)

Kurt-von-Laven:
Apologies for making the same mistake twice. I just realized the argument name is likely going to trigger CSpell for our users and should be `--remove-container`. I can open a pull request to make the change before we release it if folks agree.

nvuillam:
go @Kurt-von-Laven :)

Kurt-von-Laven:
Went for it in #1693.

PR #1565, opened 2022-06-25
Add last commit badges

bdovaz:
## Proposed Changes

Add the last commit badge in all the tables where linters, flavors, etc... are documented using: https://shields.io/category/activity

This way we can know at a glance the status of the linter repository, i.e. whether it is still maintained or abandoned and be able to make decisions whether to mark a linter as deprecated and subsequently remove it, etc.

image


nvuillam:
Wrong click.... is the merge safe ?
I'll check tomorrow at worse

bdovaz:
@nvuillam FIxed in #1566

PR #1576, opened 2022-07-05
Disable PHP Memory Limit in Docker container

stackcoder:
The default memory limit of `128M` is not enough to lint larger projects.

Since this is a development tool (running on trusted input, or in isolated environments) and not a production server its safe to disable this limit.

## Fixes:
```
PHP Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate xxxxx bytes)
```

## Proposed Changes:
1. Disable PHP memory limit

## Current Workaround:
Apply the changes by extending the Docker image with an new Dockerfile.
```Dockerfile
FROM megalinter/megalinter:latest

RUN echo 'memory_limit = -1' > /etc/php7/conf.d/02_memlimit.ini \
    && echo 'memory_limit = -1' > /etc/php8/conf.d/02_memlimit.ini
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -269,7 +269,9 @@ RUN wget --tries=5 -q -O phive.phar https://phar.io/releases/phive.phar \
     && mv phive.phar /usr/local/bin/phive \
     && rm phive.phar.asc \
     && update-alternatives --install /usr/bin/php php /usr/bin/php7 100 \
-    && update-alternatives --install /usr/bin/php php /usr/bin/php8 10
+    && update-alternatives --install /usr/bin/php php /usr/bin/php8 10 \
+    && echo 'memory_limit = -1' > /etc/php7/conf.d/02_memlimit.ini \
+    && echo 'memory_limit = -1' > /etc/php8/conf.d/02_memlimit.ini
 
 
 # POWERSHELL installation


llaville:
What PHP linter run you in trouble (PHP built-in or PHPLint ?)

I'm :-1: to apply unlimited memory usage by default. 
I suggest this alternative if you've a BIG file that consume more than 128Mo

Put into your source code a `my-php.ini` file with at least such contents 
```ini
memory_limit = ${PHP_MEMORY_LIMIT}
```

And if you want to pass a memory limit (like `256M`) or want unlimited (`-1`), depending of your linter, put in you `.mega-linter.yml` config file 

```
PHP_BUILTIN_ARGUMENTS: "-c /tmp/lint/my-php.ini"
```

PHPLinter is not yet ready to accept PHP external config file `-c` (see https://github.com/overtrue/phplint/blob/8.1/src/Linter.php#L157-L167)

And to set your `PHP_MEMORY_LIMIT` env var, call MegaLinter with `-e PHP_MEMORY_LIMIT=256M` or `-e PHP_MEMORY_LIMIT=-1` syntax

https://www.php.net/manual/en/configuration.file.php#example-27

llaville:
About PHPLint missing feature, I've just opened a report at https://github.com/overtrue/phplint/issues/137

llaville:
FYI: PHPLint has been released with new versions : 
- [3.2](https://github.com/overtrue/phplint/releases/tag/3.2.0) (PHP 5.5, 5.6 / PHP 7 compatibility)
- [4.3](https://github.com/overtrue/phplint/releases/tag/4.3.0) (PHP 8.0 compatibility)
- [5.3](https://github.com/overtrue/phplint/releases/tag/5.3.0) (PHP 8.1 compatibility)

@stackcoder If you want to contribute, please share your source code and the linter that raised your memory error (contextual data was not identified in your primary request)

stackcoder:
@llaville Sorry, I can't share the source code. Unfortunately it happened when analyzing proprietary code.

Why should a PHP linter have a memory limit, and an another one not? Just because PHP was build to serve webpages is not a valid argument here. Also I expect when system stability is required while running megalinter the way to go is limiting the container or virtual machine, not the application which is possibly processing untrusted data.

PR #1593, opened 2022-07-11
Fix `.pre-commit-hooks.yaml` link in docs

mondeja:
The link is currently raising a 404 error.

### DIFF
--- a/TEMPLATES/.jscpd.json
+++ b/TEMPLATES/.jscpd.json
@@ -10,6 +10,7 @@
     "**/.rbenv/**",
     "**/.venv/**",
     "**/report/**",
+    "**/megalinter-reports/**",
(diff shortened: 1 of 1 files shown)


nvuillam:
@Kurt-von-Laven you can merge if you like, it will be in tomorrow 's patch :)

Kurt-von-Laven:
I think GitHub stripped all of my privileges as a security precaution when we moved orgs.

nvuillam:
@Kurt-von-Laven check your mails, you have a pending invite since sunday :)

image

Kurt-von-Laven:
Whoops, thank you for straightening me out. The invite got caught by my over-aggressive filters, but I accepted it now!

PR #1657, opened 2022-07-26
Fix OUTPUT_DETAIL usage

bdovaz:
Moved the OUTPUT_DETAIL check to behave in the same way as these 2 reporters:

https://github.com/oxsecurity/megalinter/blob/main/megalinter/reporters/TextReporter.py#L20

https://github.com/oxsecurity/megalinter/blob/main/megalinter/reporters/ConsoleLinterReporter.py#L25


nvuillam:
thanks :)

PR #1701, opened 2022-08-03
Surfacing warnings during successful runs

### LINKED ISSUE #1541: Linters with warnings only
Some linters allow you to config some rules as warnings, golangci-lint is an example.  If the linters only has warnings and the process returns code is 0 the linter is marked success and the warning are lost in the linter logs. The only way to get the extra data is to place detailed report on which does for all linters and provides too much data.

I was thinking we could create a cli_lint_warning_count: regex_number and cli_lint_warnings_regex: "Total warnings found: ([0-9]+)" to the descriptor schema and linter that have warning can add those values and we can add a method to count warnings.  If warnings only exists we can provide the linter details.

If this sound like an idea you would like to consider we will do a more details response with the scope of changes.

daniel442li:
Fixes #1541.

Changed code so that the console shows a new tab called "warnings" which shows a list of warnings. Added the regex to .yaml files for both errors and warnings as a way to display the warnings
Even when there are no errors, warnings will now show in the console and be displayed
Developers can also add their own regex to each descriptor to count warnings

TODO: Add the warnings to additional reporters

## Proposed Changes

1. New schema additions, regex of warnings and warning count
2. Warning displayed on the console and other reporters in the future
3. .yaml file regex as yaml linter can very easily be counted

## Readiness Checklist
Run megalinter and there should now be a warning tab. Only .yaml files at the current moment support the counting of warnings. Other files will need additional regex to be written.
### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 Note: Can be used with `megalinter/megalinter@beta` in your GitHub Action mega-linter.yml file, or with `megalinter/megalinter:beta` docker image
 
+- Addition of warnings to reporters and logic changes to surface warnings even when there are no errors. Addition of cli_lint_warning_count & cli_lint_warning_regex variables to the JSON schema. [(#1541)](https://github.com/oxsecurity/megalinter/issues/1541)
 - Fixes about JSON Schema [(#1621)](https://github.com/oxsecurity/megalinter/issues/1621)
 - Add [checkmake](https://github.com/mrtazz/checkmake) to lint Makefile
 - Avoid cspell to lint all files. Lint only other linter files [(#1648)](https://github.com/oxsecurity/megalinter/issues/1648)
--- a/megalinter/Linter.py
+++ b/megalinter/Linter.py
@@ -111,6 +111,8 @@ class Linter:
         self.cli_lint_extra_args_after = []
         self.cli_lint_errors_count = None
         self.cli_lint_errors_regex = None
+        self.cli_lint_warnings_count = None
+        self.cli_lint_warnings_regex = None
         # Default arg name for configurations to use in linter version call
         self.cli_version_arg_name = "--version"
         self.cli_version_extra_args = []  # Extra arguments to send to cli everytime
@@ -332,6 +334,7 @@ class Linter:
             self.return_code = 0
             self.number_errors = 0
             self.total_number_errors = 0
+            self.total_number_warnings = 0
             self.number_fixed = 0
             self.files_lint_results = []
             self.start_perf = None
@@ -558,6 +561,9 @@ class Linter:
                 index = index + 1
                 return_code, stdout = self.process_linter(file)
                 file_errors_number = 0
+                file_warnings_number = 0
+                file_warnings_number = self.get_total_number_warnings(stdout)
+                self.total_number_warnings += file_warnings_number
                 if return_code > 0:
                     file_status = "error"
                     self.status = "warning" if self.disable_errors is True else "error"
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -37,6 +38,9 @@ Note: Can be used with `megalinter/megalinter@beta` in your GitHub Action mega-l
   - [checkov](https://www.checkov.io/) from 2.1.69 to **2.1.70** on 2022-07-24
   - [markdownlint](https://github.com/DavidAnson/markdownlint) from 0.32.0 to **0.32.1** on 2022-07-25
   - [checkov](https://www.checkov.io/) from 2.1.70 to **2.1.74** on 2022-07-25
+  - [bash-exec](https://tiswww.case.edu/php/chet/bash/bashtop.html) from 5.1.16 to **5.0.16** on 2022-07-26
```
nvuillam, line 41: Plz remove these lines which has been automatically added during the test classes run :)

megalinter/Linter.py
```diff
@@ -1076,6 +1090,43 @@ def get_total_number_errors(self, stdout: str):
         else:
             return 1
 
+    # Find number of warnings in linter stdout log
+    def get_total_number_warnings(self, stdout: str):
```
nvuillam, line 1094: Please could you factorize and create a method get_total_number_issues(self, level, stdout) ? 
Depending on level you would use cli_lint_errors_count or cli_lint_warnings_count, etc... :)

megalinter/Linter.py
```diff
@@ -1076,6 +1090,43 @@ def get_total_number_errors(self, stdout: str):
...
+            total_warnings = sum(int(m) for m in matches)
+        # Count all lines of the linter log
+        elif self.cli_lint_warnings_count == "total_lines":
+            total_warnings = sum(
+                not line.isspace() and line != "" for line in stdout.splitlines()
+            )
+        # Return result if found, else default value according to status
+        if total_warnings > 0:
+            return total_warnings
+        if self.cli_lint_warnings_count is not None:
```
nvuillam, line 1120: it will display a warning in log for each linter which does not have cli_lint_warnings_count
Please could you let None value, then display nothing in the warning column instead of 0 ?

nvuillam (changes requested):
Plz check comments :)

nvuillam:
That's a great initiative, I was thinking about it for some time now, thanks for implementing it :)
Please could use also update Github & Gitlab reporter classes ?
And if possible add warning properties in more descriptors than just yaml :)

daniel442li:
Yup sounds good. If you happen to know any of the regex of those other files, that would be great. I'll push another commit in the coming days!

nvuillam:
I do not (so many linters ^^) but my pattern is the following:
- copy-paste the result of a linter in https://regex101.com/
- try until I find the good regexes :D 

I used to store test classes success and error logs and display them in online doc, but since v6 it seems i've lost that... i'll try to find why and fix it ^^

Kurt-von-Laven:
[jscpd](https://github.com/kucherenko/jscpd/tree/master/fixtures) makes a convenient test repository as it contains test files of many different extensions one could run MegaLinter on to view sample linter output for all linters.

megalinter/Linter.py
daniel442li: I've been thinking about the architecture of a change like this. Because errors are found when a return_code > 0 is used, it only runs sometimes. However, because warnings can still show when the return code is 0, if we want warnings to surface we either have to seperate the warning code (what I have right now) or look for both warnings and errors even if the return_code is 0. What do you think?

megalinter/Linter.py
daniel442li: I believe the change I just made should fix this as warnings would display one sometimes based on status

PR #1765, opened 2022-08-19
Fix: Match `files_sub_directory` as a prefix

giner:
## Proposed Changes

Properly match `files_sub_directory` as a prefix instead of partial string matching.

Before

- If directory named "files_sub_directory" exists then all files
  matching "files_sub_directory" are selected. For example if
  files_sub_directory is set to "dir" and dir exists under the workspace
  directory then "mode/direct/file.yml" and "new_directory/f.yml" will
  be selected

After

- If directory named "files_sub_directory" exists then all files
  which names start with "files_sub_directory" are selected. For example if
  files_sub_directory is set to "dir" (or "./dir" or "dir/" or "./dir/")
  and dir exists under the workspace directory then "dir/file.yml" will be
  selected but not "new_directory/f.yml" and not "some/dir/f.yml"

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Note: Can be used with `oxsecurity/megalinter@beta` in your GitHub Action mega-l
 - Add Makefile linter within python flavor ([#1760](https://github.com/oxsecurity/megalinter/issues/1760))
 - Set DEFAULT_WORKSPACE as git safe directory per default [#1766](https://github.com/oxsecurity/megalinter/issues/1766)
 - Improve documentation for TAP_REPORTER
+- Fix: Properly match `files_sub_directory` as a prefix instead of partial string matching
 
 - Linter versions upgrades
   - [actionlint](https://rhysd.github.io/actionlint/) from 1.6.15 to **1.6.16**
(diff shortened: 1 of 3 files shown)


nvuillam:
it seems it does not work with some test classes using files_sub_directory...

Maybe something to play with workspace MegaLinter object attribute ?

giner:
> it seems it does not work with some test classes using files_sub_directory...
> 
> Maybe something to play with workspace MegaLinter object attribute ?

I'm trying to find an easy way to run a subset of tests locally. Any hints?

nvuillam:
Instructions are here :)

https://oxsecurity.github.io/megalinter/latest/contributing/

To locally run test classes and debug with breakpoints , you can use VsCode extension Test Explorer

image



If you use PyCharm it is native

giner:
I've seen the doc before and it was saying that I had to trigger a pipeline in any case to build a container. The only thing I could do (according to the docs) is limiting what to build (quick build) and what to test but I'd have to wait an approval every time and then the time to be built and be tested. I'll try vscode or pycharm.

nvuillam:
@giner the pipeline is indeed mandatory, but you can locally run without the docker image if you have installed the linter globally in your computer
The conditions are not the same than in CI with built image, but they are often similar so what works locally has good chances to work in CI :)

nvuillam:
I think you must solve conflicts before I can accept the CI jobs ^^

nvuillam (approved):
Good catch, thanks a lot for your contribution :)

PR #1770, opened 2022-08-20
Add DEFAULT_WORKSPACE path to git safe.directory by default

### LINKED ISSUE #1766: Add git "safe directory" for tekton convention "/workspace/source" dockerfile entrypoint (like github_workspace) or generalize.
When running megalinter docker in a tekton pipeline, one can stumble upon the quite recent git safe.directory problem so one has to add "git config --global --add safe.directory PATH". I see like you already added patch specifically for github_workspaces default path in entrypoint.sh. Note: An convention (many tekton examples shows this, so I guess other people than us will also use this path) is to use /workspace/source path, even if one does not have to of course.

Either just add an extra setting also for TEKTON_WORKSPACE, in the entrypoint.sh, and set /workspace/source/ (fastest fix) or maybe even generalize it to handle all future cases including github,  with something like
 if [[ -d "${DEFAULT_WORKSPACE}" ]] ; then git config --global --add safe.directory "${DEFAULT_WORKSPACE}"; echo "Setting git safe dir ${DEFAULT_WORKSPACE}" ;~                              |     else echo 'no dir' ; fi 

Further, if generic maybe make the setting configurable - Maybe an env to "SET_GIT_WORKSPACE_AS_GIT_SAFE_DIR" , or something like that.

We create our own image , so we can add this on our own, but,, if we have the problem, other surely has too. Would be nice to have "out of the box" and spare the enduser some possible error hunt.

janderssonse:
Fixes #1766 

Adds git safe.directory DEFAULT_WORKSPACE by default in entrypoint.sh.

## Proposed Changes

See discussion and motivation in [#1766 ](https://github.com/oxsecurity/megalinter/issues/1766).

Note: I made the option default, but left an small undocumented env variable to let anyone really wishing the old way configure it by setting WORKSPACE_AS_SAFE_DIR to "false". If you don't get any questions about this default behaviour over time, I would suggest to remove that flag. On the other hand,  If you do get questions for anyone wishing this, let's document WORKSPACE_AS_SAFE_DIR properly and add it to the documentation.

### Author/Contributor

### Reviewing Maintainer

Signed-off-by: Josef Andersson <josef.andersson@gmail.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 Note: Can be used with `oxsecurity/megalinter@beta` in your GitHub Action mega-linter.yml file, or with `oxsecurity/megalinter:beta` docker image
 
+- Set DEFAULT_WORKSPACE as git safe directory per default [#1766](https://github.com/oxsecurity/megalinter/issues/1766)
 - Update pre-commit hooks from v5 to v6 ([#1755](https://github.com/oxsecurity/megalinter/issues/1755)).
 - Fix version in URL in logs produced by reporters
 - Improve documentation for TAP_REPORTER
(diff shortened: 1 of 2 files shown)


nvuillam (approved):
Nice, thanks for your contribution :)

PR #1782, opened 2022-08-22
Match regex without workspace and sub_directory

giner:
## Proposed Changes

This change allows to use safer regular expression with "beginning of string"
anchor, i.e. `^this_is_a_beginning_of_string` for filtering files

Before

- Regex matching for including and excluding files is done against
  the whole path with workspace and sub_directory, i.e. if the file
  path is `workspace/sub_directory/anotherdir/somefile.yml`, then adding
  "direct" or "work" to EXCLUDE or INCLUDE will affect this file.

After

- Regex matching for including and excluding files is done against
  normalized path to file without workspace and sub_directory, i.e. if the file
  path is `workspace/sub_directory/anotherdir/somefile.yml`, then adding
  "direct" or "work" to EXCLUDE or INCLUDE won't affect this file however
  adding "^another" will.

### Author/Contributor

### Reviewing Maintainer


nvuillam:
This seems nice, but... I'm afraid of regressions :o 

Are we really sure that no one uses full path for regex filtering variables ?

What is your case that requires such enhancement ? ^^

giner:
> Are we really sure that no one uses full path for regex filtering variables ?

We are not. Someone could use expressions starting with a shash or starting from a sub directory. I could work around both (the first one by pretending `./` and the second one by excluding only prefix but not subdirectory) however it's a trade off about when to "suffer" now or in the future.

> What is your case that requires such enhancement ? ^^

The attempt to anchor the beginning of filenames so that the expressions don't match wrong files or directories accidentally due to a partial match. Let's say I want to match `dir/xxx` but not `dir2/dir/xxx` and not "dir" in a prefix then before this change 8s applied `/dir/` would match all three and after this change I could property anchor it with `^dir/`.

nvuillam (approved):
mmmm sometimes we have to live with some risks :)
Ok let's go ! :)

PR #1820, opened 2022-09-01
Add cli_lint_errors_count to java descriptor

ansraliant:
In order to parse the amount of errors Checkstyle finds, we need to tell the parser which type of message the linter responds with.
In case of checkstyle, it has a message like Checkstyle ends with ([0-9]+) errors which corresponds to `regex_number`. Otherwise megalinter will respond always with 1.

Fixes #
Tries to fix an issue when the count of errors reported by checkstyle are not properly parsed.

## Proposed Changes

1. adds cli_lint_errors_count

### Author/Contributor

### Reviewing Maintainer


nvuillam (approved):
Good catch, thanks a lot for your contribution :)

PR #1862, opened 2022-09-11
fix: correct cli arguments for terrascan

ahmadnassri:
terrascan uses `--config-path` _(currently set to use `--config`)_

image


nvuillam (approved):
Good catch, thanks :)

PR #1897, opened 2022-09-21
Update Devcontainer files to use python base image

omusavi:
The current devcontainer files used the main Dockerfile that included every linter installed. For developing within the MegaLinter environment, that is typically overkill. Instead, using the python3 base image allows for the dev environment to be spun up quickly. It also includes some QOL features like built-in ZSH and Oh-My-Zsh for a great terminal experience.

## Proposed Changes

1. Create a new Dockerfile that is only used by the development environment
2. Point the .devcontainer.json file to the new Dockerfile

### Author/Contributor

### Reviewing Maintainer


.devcontainer/devcontainer.json
```diff
@@ -1,9 +1,56 @@
...
+				"python.linting.pydocstylePath": "/usr/local/py-utils/bin/pydocstyle",
+				"python.linting.pylintPath": "/usr/local/py-utils/bin/pylint"
+			},
+			
+			// Add the IDs of extensions you want installed when the container is created.
+			"extensions": [
+				"ms-python.python",
+				"ms-python.vscode-pylance",
+				"github.vscode-pull-request-github",
+				"rogalmic.bash-debug"
```
nvuillam, line 43: any chance you could add here the recommended extensions of the repo ? :)


https://github.com/oxsecurity/megalinter/blob/main/.vscode/extensions.json

nvuillam (approved):
@omusavi  that's great, many thanks for the contribution :)

I validate this PR because it's already a great progression in the MegaLinter codespaces capability ^^

Next steps (can be done by you, me, or anyone willing ^^):
- Update Contributing documentation
- I see python 3.9 related lint warnings -> let's use the same python version than ML ^^
- Merge RUNs instructions to reduce docker layers
- Add some linters pre-installed anyway to have more ready-to-use test cases

omusavi:
@nvuillam all those sound like great additions. Sorry I got the Python version wrong, that should be a quick fix, I will submit that today and if I have time I can add some notes in the Contributing guide as well.

WRT the already built image - yes it is possible and is not too difficult. You can see in [this commit](https://github.com/omusavi/megalinter/commit/5c9b655ad65522aad91f73592ac7a7ff4ba0c147) the necessary work to convert from the existing Microsoft managed dev container to one with any base image you want: https://github.com/omusavi/megalinter/commit/5c9b655ad65522aad91f73592ac7a7ff4ba0c147.

In the commit above, I used the megalinter:beta as the base image. Since that image uses Alpine, I also added a special script file that gets run on the docker image which sets it up to be able to communicate with the devcontainer (things like setting up the user and some tooling). Overall not too much work.

The main difference right now between using the megalinter:beta image and the current implementation is time to load the codespace. The megalinter image took about 3-4 minutes to start the codespace the first time whereas the python image in this PR took around 30 seconds. For people looking to contribute, I would think the faster we can get the dev environment up and running the better. If we wanted to use a scaled down base image, something in between the beta image and the current image, that would probably be the best user experience (in my opinion)

nvuillam:
I was just thinking about publishing the image corresponding to the Dockerfile you built :)

omusavi:
@nvuillam then you wouldnt need to do any additional work with the scripts and such!

PR #1900, opened 2022-09-22
Set Devcontainer Python version to 3.10

omusavi:
In the previous PR #1897 , the `.devcontainer.json` file was passing in Python 3.9 instead of 3.10, which is the version Megalinter is built on. This PR fixes the version error.

## Proposed Changes

1. Update version in `.devcontainer.json` to 3.10

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -4,7 +4,7 @@
 		"dockerfile": "Dockerfile",
 		"context": "..",
 		"args": { 
-			"VARIANT": "3.9",
+			"VARIANT": "3.10",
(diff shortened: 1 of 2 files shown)


nvuillam:
Great, thanks :)
Is it possible to specify the version revision ? 

For example MegaLinter is based on python 3.10.4

nvuillam (approved):

omusavi:
@nvuillam unfortunately not with the microsoft images: https://mcr.microsoft.com/v2/vscode/devcontainers/python/tags/list

We would need to create our own base image to lock it do a revision version

PR #1934, opened 2022-10-05
`markdownlint-cli` → `markdownlint-cli2`

parkerbxyz:
## Proposed Changes

Replace `markdownlint-cli` with `markdownlint-cli2`.

**Motivation:** MegaLinter doesn't currently play nice with directory-specific markdownlint configuration files. Using `markdownlint-cli2` in place of `markdownlint-cli` will fix this.

**Advantages:**

- Speed (allegedly)
- Configuration files can live anywhere and automatically apply to their part of the directory tree
  The markdownlint VS Code extension is powered by `markdownlint-cli2`, so everything should _just work_ without special configuration when using the extension in conjunction with MegaLinter.
- Fewer dependencies

Additional information: https://dlaa.me/blog/post/markdownlintcli2

### Author/Contributor

### Reviewing Maintainer


parkerbxyz:
The only markdownlint test that I see failing is this one:

```
megalinter/tests/test_megalinter/linters/markdown_markdownlint_test.py::markdown_markdownlint_test::test_failure FAILED [ 46%]
```

I haven't been able to find any logs indicating why it's failing and I haven't figured out how to run the tests locally to debug further. Any help here would be greatly appreciated. 🙂

nvuillam:
Maybe you can start y what are the advatages og megalinter to use maskdownlint-cli2 instead of markdownlint ? :)

parkerbxyz:
> Maybe you can start y what are the advatages og megalinter to use maskdownlint-cli2 instead of markdownlint ? :)

I'm not sure I understand the question... If you're asking what the advantages of `markdownlint-cli2` are over the original markdownlint CLI, I addressed that in [the PR description](https://github.com/oxsecurity/megalinter/pull/1934#issue-1397262576). Let me know if there's something I should elaborate on there or if you have any other questions. 🙂

Kurt-von-Laven:
> The only markdownlint test that I see failing is this one:
> 
> ```
> megalinter/tests/test_megalinter/linters/markdown_markdownlint_test.py::markdown_markdownlint_test::test_failure FAILED [ 46%]
> ```
> 
> I haven't been able to find any logs indicating why it's failing and I haven't figured out how to run the tests locally to debug further. Any help here would be greatly appreciated. 🙂

I think this is very valid feedback. I am inclined to give our CI pipeline some love if @nvuillam agrees. I am open to suggestions on where to begin. Otherwise, I will just try to keep the PRs small in order to solicit feedback.

nvuillam:
image

There are errors in _bad_ testing files
They are not detected by markdownlint-cli2, that's why CI job fails

parkerbxyz:
It looks like the test is adding `markdownlint-cli2` as an argument to the command, like so:

```sh
markdownlint-cli2 markdownlint-cli2 /tmp/lint/.automation/test/markdown/markdown_bad_1.md /tmp/lint/.automation/test/markdown/markdown_bad_2.md
```

This appears to be because of how the container works. More on that [here](https://github.com/DavidAnson/markdownlint-cli2#container-image). If we want to use the container, we'll need to use the `--entrypoint` flag to indicate which command to run without adding it twice.

I'm going to see if the build success without the container and go from there.

nvuillam:
I think using npm package could be ok :)

parkerbxyz:
```txt
megalinter/tests/test_megalinter/helpers/utilstest.py:604: in assert_file_has_been_updated
    test_self.assertTrue(updated, f"{file_name} has been updated")
E   AssertionError: False is not true : markdown_for_fixes_1.md has been updated
```

If I run the following command locally on this branch, 17 files, including `markdown_for_fixes_1.md` get fixed:

```sh
npx mega-linter-runner --path .automation/test/sample_project_fixes/ --fix
```

Any ideas why it might behave differently in the test?

nvuillam:
image

I'm less and less confident of ascending compatibility when switching from markdownlint-cli to markdownlint-cli-2... having different executables for fixing and "config" is an unique pattern I saw i all linters, so updating MegaLinter core just for one linter, which does apparently the same than the one it's supposed to replace...  I'll need much more convincing that a link to a PR ^^

parkerbxyz:
> I'll need much more convincing that a link to a PR ^^

What link to a PR are you referring to? As mentioned in [the PR description](https://github.com/oxsecurity/megalinter/pull/1934#issue-1397262576), my motivation here is to add support for directory-specific markdownlint configuration files to bring the linter in line with how the official VS Code extension works. This would allow the configuration files to _just work_ without needing to have MegaLinter-specific markdownlint configuration options/files.

nvuillam:
I'm ok for the switch to have better perfs and folder local config... but less ok to have to make MegaLinter core even more complicated ^^
markdownlint-cli2 can not work with a single executable + arguments ? :)

parkerbxyz:
> markdownlint-cli2 can not work with a single executable + arguments ? :)

Unfortunately not. The only arguments accepted are globs (files/directories):

```
Syntax: markdownlint-cli2 glob0 [glob1] [...] [globN]
        markdownlint-cli2-fix glob0 [glob1] [...] [globN]
        markdownlint-cli2-config config-file glob0 [glob1] [...] [globN]
```

nvuillam:
And what does the markdownlint-cli2-config ?

nvuillam:
Maybe the solution could be to keep markdownlint-cli and add markdownlint-cli2 as a new linter :) markdownlint-cli would remain by default but with a variable we could opt to markdownlint-cli2, like we do for prettier and standard for js ^^

parkerbxyz:
> Maybe the solution could be to keep markdownlint-cli and add markdownlint-cli2 as a new linter :) markdownlint-cli would remain by default but with a variable we could opt to markdownlint-cli2, like we do for prettier and standard for js ^^

That seems reasonable. We'll still need to get `markdownlint-cli2` working within MegaLinter in that case.

With the latest dev build, I'm back to this error:

> ```
> megalinter/tests/test_megalinter/helpers/utilstest.py:604: in assert_file_has_been_updated
>     test_self.assertTrue(updated, f"{file_name} has been updated")
> E   AssertionError: False is not true : markdown_for_fixes_1.md has been updated
> ```
> 
> If I run the following command locally on this branch, 17 files, including `markdown_for_fixes_1.md` get updated/fixed:
> 
> ```shell
> npx mega-linter-runner --path .automation/test/sample_project_fixes/ --fix
> ```
> 
> Any ideas why it might behave differently in the test?

parkerbxyz:
Checks are passing. Ready for review!

echoix:
About the changes in the configuration schema. What happens to existing users, their config is no longer valid?

Since from what I know, the jsonschema store isn't versioned, how does existing releases play with that?

I'm asking questions, but maybe they don't need to be "solved" for the PR to go through. Only _understand_ the implications

parkerbxyz:
> About the changes in the configuration schema. What happens to existing users, their config is no longer valid?
> 
> Since from what I know, the jsonschema store isn't versioned, how does existing releases play with that?
> 
> I'm asking questions, but maybe they don't need to be "solved" for the PR to go through. Only _understand_ the implications

Great questions! I'm actually not sure. Perhaps this should be considered a breaking change with a note that configurations with `MARKDOWN_MARKDOWNLINT_CLI` should be changed to `MARKDOWN_MARKDOWNLINT_CLI2`? Alternatively, I could revert the changes to the schema, or maybe there's a way both options could be supported.

echoix:
Just like this, does markdownlint-cli2 support the same config and options that markdownlint-cli does? (In other words, if it were a mathematical set, the cli2 contains all of cli, plus cli2-specific elements). If so, we could just say that we substitute the binary that actually does the linting, but keep the variable names of ...-cli intact.

Does it make sense?

parkerbxyz:
From the [markdownlint-cli2 README](https://github.com/DavidAnson/markdownlint-cli2?tab=readme-ov-file#overview):

> - `markdownlint-cli2` is configuration-based and prioritizes speed and simplicity.
> - `markdownlint-cli2` supports all the features of markdownlint-cli (sometimes a little differently).

> ## Compatibility
> 
> ### `markdownlint-cli`
> 
> - The glob implementation and handling of pattern matching is slightly
>   different.
> - Configuration files are supported in every directory (vs. only one at the
>   root).
> - The `INI` config format, `.markdownlintrc`, and `.markdownlintignore` are not
>   supported.



The `--config`, `--fix`, and `--help` parameters appear to be the same (though cli2 doesn't appear to support `-c`, `-f`, and `-h`). `markdownlint-cli` supports additional parameters.

- [markdownlint-cli usage](https://github.com/igorshubovych/markdownlint-cli#usage)
- [markdownlint-cli2 usage](https://github.com/DavidAnson/markdownlint-cli2/blob/main/README.md#command-line)

echoix:
We're planning and have already some work merged for v9, a major version. @nvuillam what is your opinion on what we accept as changes to users?

echoix:
This PR has been open for way too long

nvuillam:
@echoix v9 is the next version (no ETA yet as we need to implement the custom flavors), so we can add breaking changes as long as we document them :)

parkerbxyz:
Should I continue keeping this PR up-to-date by resolving conflicts as they arise?

PR #1944, opened 2022-10-07
New linter: logstash config

Yann-J:
This adds a linter for Logstash configuration files, using [mustache](https://github.com/breml/logstash-config).


megalinter/descriptors/logstash.megalinter-descriptor.yml
```diff
@@ -0,0 +1,32 @@
+descriptor_id: LOGSTASH
+descriptor_type: tooling_format
+file_extensions:
+  - ".conf"
```
echoix, line 4: I'm concerned here that having this descriptor enabled in a vanilla config with the full megalinter might run on unexpected files. `.conf` isn't an extension unique to logstash. 

Will the linter, if ran on .conf files that aren't logstash config files, but any other config file, will throw false positives for everyone else?

Yann-J: Hmm good question, it will probably throw indeed... It's a fairly specific linter so indeed should probably be opt-in only, but not sure how that can be handled?

Yann-J: By convention, Logstash conf is often under a `**/pipeline` directory, so maybe we can filter on this, but not sure how I can specify this?

llaville: > By convention, Logstash conf is often under a `**/pipeline` directory, so maybe we can filter on this, but not sure how I can specify this?

You can try with syntax like used in `actionlint` linter (see https://github.com/oxsecurity/megalinter/blob/main/megalinter/descriptors/action.megalinter-descriptor.yml#L5)

Yann-J: OK, I've added this! I've set it to `logstash` by default, which I guess may not generally be appropriate, but at least should avoid conflicts with other formats by default, and can be overridden to suit specific cases.

Yann-J:
Hey just a ping here @echoix , is there anything else I can do to improve my PR?

nvuillam:
@Yann-J as this linter seems not widely adopted, and not very maintained also... maybe I can suggest to provide it as a MegaLinter plugin instead of adding it in Core MegaLinter ?

image

Yann-J:
Yeah, that's fair, I kind of suspected as much. We already have it running as a plugin right now, but it's not yet published... will update here when it is...

nvuillam:
@Yann-J You can add links in [plugins README section](https://github.com/oxsecurity/megalinter#plugins-catalog), and i'll be glad to make an announcement about it in the next MegaLinter release :)

Yann-J:
ok, closing in favor of #2015

PR #1986, opened 2022-10-21
Replace `set-output` command

parkerbxyz:
Replaces the soon-to-be deprecated `set-output` command.

Reference: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### Author/Contributor

### Reviewing Maintainer


parkerbxyz:
The build no longer shows the following annotations, so it looks like these changes are working as intended! 🎉 

> #### [Tests + Deploy Docker Image - DEV](https://github.com/oxsecurity/megalinter/actions/runs/3300992229/jobs/5446029988#step:5:11)
> The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

nvuillam:
@parkerbxyz please can you make your branch up to date with main ? it will fix CI issues not related to your PR :)

nvuillam (approved):
Thanks for the catch :)
Let's hope you know what you are doing, else release CI jobs are broken :D :D

.automation/build.py
```diff
@@ -2757,7 +2757,7 @@ def manage_output_variables():
                 updated_versions = 1
                 break
         if updated_versions == 1:
-            print("::set-output name=has_updated_versions::1")
+            print('"has_updated_versions=1" >>"$GITHUB_OUTPUT"')
```
Kurt-von-Laven, line 2760: The Bash changes look correct at a glance, but I suspect printing Bash syntax from Python has no special effect. See EnricoMi/publish-unit-test-result-action#360 for an example of the correct syntax.

Kurt-von-Laven (commented):
@parkerbxyz, thanks for taking a pass at this.

.automation/build.py
nvuillam: I'm afraid you're right... I'll rollback the PR to not break behaviour until we have a working solution ( cc  @parkerbxyz  )

parkerbxyz: > I suspect printing Bash syntax from Python has no special effect

In that case, is it working as expected without my changes?

How about something like this? 

```python
import shlex, subprocess

add_to_output_cmd = 'echo "has_updated_versions=1" >>"$GITHUB_OUTPUT"'
add_to_output_args = shlex.split(add_to_output_cmd)
subprocess.run(add_to_output_args)
```

```suggestion
            add_to_output_cmd = 'echo "has_updated_versions=1" >>"$GITHUB_OUTPUT"'
            add_to_output_args = shlex.split(add_to_output_cmd)
            subprocess.run(add_to_output_args)
```

megalinter/MegaLinter.py
```diff
@@ -822,7 +822,7 @@ def display_header():
         logging.info("")
 
     def check_results(self):
-        print(f"::set-output name=has_updated_sources::{str(self.has_updated_sources)}")
+        print(f'"has_updated_sources={str(self.has_updated_sources)}" >>"$GITHUB_OUTPUT"')
```
parkerbxyz, line 825:
```suggestion
        add_to_output_cmd = f'echo "has_updated_sources={str(has_updated_sources)}" >>"$GITHUB_OUTPUT"'
        add_to_output_args = shlex.split(add_to_output_cmd)
        subprocess.run(add_to_output_args)
```

nvuillam: It could work, but it will update $GITHUB_OUTPUT in docker image
Will Github Action workflow retrieve the variable from inside the docker image ? if not we have to find another way (maybe some dummy file in artifacts that we could check for existence ...

Kurt-von-Laven: Yes, the existing code looks correct, just deprecated. The goal here is simply to append to the file named `GITHUB_OUTPUT`, so we can achieve that without spawning a process.

```py
from os import environ

output_file = environ["GITHUB_OUTPUT"]
with open(output_file, "a", encoding="utf-8") as output_stream:
    print(f"has_updated_sources={str(self.has_updated_sources)}", file=output_stream)
```

nvuillam: @Kurt-von-Laven print(), not output_stream.write() ?

Kurt-von-Laven: The last line could be replaced with:

```py
output_stream.write(f"has_updated_sources={str(self.has_updated_sources)}\n")
```

Any preference?

Just saw your previous message. To be clear, we don’t want to overwrite the environment variable `GITHUB_OUTPUT`, which is essentially the path to a magic dummy file that gets sourced at the beginning of each subsequent step.

nvuillam: I'd prefer the write option , it's more readable that a print that we expect more to be just a line in the console :)

nvuillam: @parkerbxyz maybe u'd like to ressuscitate your PR and update it ? :) ( ps: just been rickrolled... 😠 🤣 )

Kurt-von-Laven: Sounds good to me. I updated the code snippet accordingly.

Ha ha, the meme that will never die.

Issue #1472, opened 2022-05-13
Move GitHub Action image to GitHub Packages


estahn:
**Is your feature request related to a problem? Please describe.**

It would be great to load the image from GitHub Packages, as I assume this will be a substantially faster pulling experience.

https://github.com/megalinter/megalinter/blob/0126180f621bdb4f84b1864e6788daba8c7b1642/action.yml#L10

**Describe the solution you'd like**

* Upload said image to GitHub Packages
* Use it in `action.yml`

Kurt-von-Laven:
We would probably need to offer the image via both Docker and GitHub Packages at least for now since it seems like a pretty breaking change. I may be missing something since I have never used GitHub Packages, but If you or anyone else finds time to open a pull request, it seems like a low-effort, high-reward way to boost pull performance for GitHub users without affecting others.

Kurt-von-Laven:
In the meantime, those using GitHub Actions might try the dirt simple [docker-cache](https://github.com/ScribeMD/docker-cache) action I wrote to improve our MegaLinter pull times.

estahn:
@Kurt-von-Laven Seems like this is already in place, not sure why its not executed:

https://github.com/megalinter/megalinter/blob/f958756b80b7a1bbc8b0115abf80a53a55ee1c70/.github/workflows/deploy-PROD.yml#L94-L104

Kurt-von-Laven:
Oh, maybe my assumption that we don’t already do this was wrong in that case. I do notice a few errors in [the logs](https://github.com/megalinter/megalinter/runs/6435661494?check_suite_focus=true) about some Docker images not being found, but they do claim a successful upload to the GitHub Container Registry. I am unfamiliar with all of these tools and on mobile right now, so please let me know if I am overlooking the obvious.

nvuillam:
The main docker image is uploaded to github registry, but not the flavor images yet

nvuillam:
I think this was because github registry allowed only one image name by repo but maybe that changed, i can have a look

Kurt-von-Laven:
I read [this blog post](https://www.google.com/amp/s/blog.atomist.com/github-container-registry/amp/) to say that they now allow multiple images per repo:

[...]

PR #2117, opened 2022-12-01
Deploy additional Docker images to GitHub Container Registry

lars-reimann:
Fixes partially #1472

## Proposed Changes

1. Deploy additional Docker images to GitHub Container Registry. This would later allow [action.yml](https://github.com/oxsecurity/megalinter/blob/main/action.yml) to pull the MegaLinter image from GitHub instead of Docker Hub, which _should_ speed up the pull phase of the action.

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/.github/linters/.cspell.json
+++ b/.github/linters/.cspell.json
@@ -1080,6 +1080,7 @@
         "regexpp",
         "rego",
         "regsub",
+        "reimann",
         "reindent",
         "relaxng",
         "remarkrc",
--- a/.github/workflows/deploy-ALPHA-flavors.yml
+++ b/.github/workflows/deploy-ALPHA-flavors.yml
@@ -78,9 +78,9 @@ jobs:
       - name: Checkout Code
         uses: actions/checkout@v3
 
-      #####################
-      # Run Deploy script #
-      #####################
+      ###################################
+      # Run Deploy script for Dockerhub #
+      ###################################
       - name: Deploy latest image to DockerHub
         env:
           # Set the Env Vars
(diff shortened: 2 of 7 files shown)


lars-reimann:
Some questions about this:

1. Should other `deploy` workflows also be updated? This includes
    * deploy-ALPHA-flavors.yml
    * deploy-BETA-flavors.yml
    * deploy-BETA-linters.yml
    * deploy-DEV.yml
    * deploy-DEV-linters.yml
    * deploy-RELEASE-linters.yml
2. Should this PR also update `action.yml` or should this be done in a separate PR?
3. How can I test these changes?

Once these questions are answered, I'll update the `CHANGELOG` accordingly.

bdovaz:
Ping @nvuillam, can you answer this questions?

nvuillam:
@bdovaz beta, beta-flavors and beta-linters will validate that it work, so the release jobs have more chances to work too :)

lars-reimann:
I recently found [this discussion](https://github.com/orgs/community/discussions/27080) claiming that GHCR is no faster than Docker for GitHub Actions. So, I wonder whether it makes sense to further pursue this PR.

nvuillam:
docker pulls on docker hub has download limitations (not applied on github actions), so it still can be iteresting to use ghcr image, especially for self hossted runners :)

Kurt-von-Laven:
Thai Pangsakulyanont [reported improved performance building and pushing to GHCR](https://dev.to/dtinth/caching-docker-builds-in-github-actions-which-approach-is-the-fastest-a-research-18ei), which makes me skeptical that pulling is somehow slower. I wrote [ScribeMD/docker-cache](https://github.com/ScribeMD/docker-cache), automating approach 📦 in the blog post, and it outperforms pulling directly from Docker Hub. In any case, we should measure the performance ourselves since it may have changed over time.

echoix:
And what about simply when there are some outages? Its native to GitHub, and the whole build doesn't need to be done twice anyway, only two pushes instead of one.

lars-reimann:
I can continue with this next week. If someone else wants to work on it in the meantime, feel free to do so.

nvuillam (approved):
Great PR, nice job, thanks a lot for your contribution :)

nvuillam:
When i try to merge i have an error :/

echoix:
> When i try to merge i have an error :/
> 
> 

Did you try on web instead of app? I can't imagine why an OAuth app is involved in that. Of course I never saw the settings on admin side, but changes to workflows have been done even during the holidays without problems.

Kurt-von-Laven:
By default, bots can't edit `.github/workflows` for security reasons (c.f., [orgs/community#25222](https://github.com/orgs/community/discussions/25222), [orgs/community#26164](https://github.com/orgs/community/discussions/26164)). In this case, since the error is coming from an OAuth App, it's possible to grant it the [`workflow` scope](https://docs.github.com/en/developers/apps/building-oauth-apps/scopes-for-oauth-apps#available-scopes) if desired. I don't know which OAuth App this is though, so I can't comment on whether granting it the ability to execute arbitrary code in our name would be wise.

nvuillam:
@Kurt-von-Laven @echoix  it worked thru web UI :)

nvuillam:
thanks again @lars-reimann :)

Issue #2006, opened 2022-10-25
DEFAULT_WORKSPACE is prefixed to SARIF-output field uri but should not be.


janderssonse:
**Describe the bug**
DEFAULT_WORKSPACE is prefixed to uri in the SARIF-output, in my case, breaks SARIF-Viewer/VSCode, but would be the same for any alike setup.

**To Reproduce**
Example: 

Run the megalinter container, here standing in the root of the repo you want to check: 

1. podman run --volume $(pwd):/repo DEFAULT_WORKSPACE=/repo -e LOG_LEVEL=INFO oxsecurity/megalinter:v6.13.0
2. Open up your fav editor, for example VSCode, (with SARIF-Vieever extension installed)
3. Open a sarif file from VSCode.(from megalinter-reports/sarif
4. While rules are shown, notice VSCode cant find the path to the error. Look in the SARIF-FILE, see that all "uri"s have the DEFAULT_WORKSPACE prefixed to them., EG /repo/src/main etc. 
5. Removing the prefixed DEFAULT_WORKSPACE so we have expected src/main. Now it works!

**Expected behavior**
DEFAULT_WORKSPACE should not be prefixed to SARIF-output. This variable is something that most likely only exists during runtime generation, and makes the SARIF-output closely tied to that run.

nvuillam:
There is already a replacement in generated SARIF for some values like /tmp/lint or /github/workspace 

The case with custom value is indeed not taken in account, we can arrange that ,:)

Meanwhile if you use /tmp/lint it will be ok :)

As some linters directly generate a SARIF file, content of SARIF folder may not be post-processed, you may directly use the unique SARIF file generated at report root :)

janderssonse:
I will give this one a shot when I have the time. I had a look at the code, and I think this one could be implemented in the method fix_sarif_physical_location by just looking if the uri field starts with the same as DEFAULT_WORKSPACE, (if set) and if so, remove that prefix.

janderssonse:
A small PR is now here https://github.com/oxsecurity/megalinter/pull/2119.
It is to be considerad as a partial fix.

That will fix the problem for the aggregated sarifreport.

[...]

PR #2119, opened 2022-12-03
Remove def.ws prefix from aggregate sarif report

janderssonse:
Fixes #https://github.com/oxsecurity/megalinter/issues/2006
(partially, see discussion there)

This will replace the default workspace prefix
from when creating the aggregate sarif report output. (def. megalinter-reports.sarif).

In other words, the artifact uri will be

/root/path/of/source/code

instead of as of now:

DEFAULT_WORKSPACE/root/path/of/source/code.

## Proposed Changes

As the DEFAULT_WORKSPACE might only exist during
runtime lints, it needs to be removed to offer a deterministic uri, for the reports.

### Author/Contributor

### Reviewing Maintainer

Signed-off-by: Josef Andersson <josef.andersson@gmail.com>

### DIFF
--- a/megalinter/reporters/SarifReporter.py
+++ b/megalinter/reporters/SarifReporter.py
@@ -221,11 +221,22 @@ class SarifReporter(Reporter):
                 # Update run in full list
                 linter_sarif_obj["runs"][id_run] = run
         return linter_sarif_obj
+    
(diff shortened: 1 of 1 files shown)


nvuillam:
@janderssonse if you update your branch to be up to date with main it should be ok

nvuillam (approved):
Great, thanks for your contribution:)

Issue #2146, opened 2022-12-19
Why does the EXTENDS value have to be a url?


bdovaz:
I want in the same directory to have:

Base configuration:

.megalinter-base.yml

Both files use base configuration:

.megalinter-use-case1.yml
.megalinter-use-case2.yml

The 3 files are in the same directory but in none of the last 2 I can put:

EXTENDS: .megalinter-base.yml

I have to put a url necessarily.

Why?

nvuillam:
`EXTENDS` property has been created to share a base MegaLinter configuration between repos :)

What is the use case that requires `EXTENDS` to target a local file ?

If you really need it, you can make a PR and add a condition `if extends_item.startswith("http"): .... else: ....` here :) https://github.com/oxsecurity/megalinter/blob/c13e920cea92c5276c1d25befe1d5b9561819c84/megalinter/config.py#L79

bdovaz:
I have created a PR

My use case is that I want to have several Azure pipelines, each one with a different execution of megalinter due to the fact that there are some linters that take a long time to execute (the REPOSITORY ones that have lint_mode project) and I want to have a lighter .megalinter.yml and another complete one that executes at night. The 2 share practically everything so I want to use EXTENDS and repeat as little as possible.

nvuillam:
That's indeed a good use case !

nvuillam:
Fixed by #2151

PR #2151, opened 2022-12-20
Allow local files with EXTENDS

bdovaz:
https://github.com/oxsecurity/megalinter/issues/2146

### DIFF
--- a/megalinter/config.py
+++ b/megalinter/config.py
@@ -76,11 +76,14 @@ def init_config(workspace=None):
         if isinstance(extends, str):
             extends = extends.split(",")
         for extends_item in extends:
-            r = requests.get(extends_item, allow_redirects=True)
-            assert (
-                r.status_code == 200
-            ), f"Unable to retrieve EXTENDS config file {config_file_name}"
-            extends_config_data = yaml.safe_load(r.content)
+            if extends_item.startswith("http"):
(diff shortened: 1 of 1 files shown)


megalinter/config.py
```diff
@@ -76,11 +76,14 @@ def init_config(workspace=None):
...
-            ), f"Unable to retrieve EXTENDS config file {config_file_name}"
-            extends_config_data = yaml.safe_load(r.content)
+            if extends_item.startswith("http"):
+                r = requests.get(extends_item, allow_redirects=True)
+                assert (
+                    r.status_code == 200
+                ), f"Unable to retrieve EXTENDS config file {config_file_name}"
+                extends_config_data = yaml.safe_load(r.content)
+            else:
+                extends_config_data = yaml.safe_load(extends_item)
```
nvuillam, line 86: You need to open the file :)

```python
with open(extends_item, "r", encoding="utf-8") as f:
    extends_config_data = yaml.safe_load(f)
```

nvuillam (changes requested):
Plz check comments ^^

megalinter/config.py
bdovaz: Done

nvuillam (approved):
Great, thanks for your contribution !

PR #2154, opened 2022-12-20
Add Trivy config file parameters

bdovaz:
- Add config_file_name and cli_config_arg_name
- Change configuration url to a more suitable one
- Change old urls to the current version
- Fix broken banner image url

### DIFF
--- a/megalinter/descriptors/repository.megalinter-descriptor.yml
+++ b/megalinter/descriptors/repository.megalinter-descriptor.yml
@@ -357,13 +357,13 @@ linters:
       - cupcake
       - security
     linter_text: |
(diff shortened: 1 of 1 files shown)


bdovaz:
I have doubts whether to put /v0.35/ or /latest/.

v0.35:
- Difficult to maintain, it is easy to forget to update the version....
- Aim for the version that corresponds to the one in the image.

latest:
- Not to be maintained.
- It always points to the latest version and it doesn't have to be the one in the image.

nvuillam:
@bdovaz I don't remember why we did not use latest
If it works with latest (and returns correctly the version number), go with latest , that's the default way for other linters

bdovaz:
@nvuillam changed!

nvuillam (approved):
Great, many thanks for your contribution :)

PR #2156, opened 2022-12-21
Use warning emoji in reporter

ashokm:
Adds warning emoji to keep this consistent with success and error reporting.


nvuillam (approved):
Nice catch, thanks for your contribution :)

PR #2150, opened 2022-12-20
Add npm-package-json-lint

bdovaz:
https://github.com/oxsecurity/megalinter/issues/2092

PR #2163, opened 2022-12-21
Add file_names_regex linter schema

bdovaz:
https://github.com/oxsecurity/megalinter/pull/2150#discussion_r1054827889

### DIFF
--- a/megalinter/descriptors/schemas/megalinter-descriptor.jsonschema.json
+++ b/megalinter/descriptors/schemas/megalinter-descriptor.jsonschema.json
@@ -815,6 +815,22 @@
             "title": "Filter on end of file name",
             "type": "array"
           },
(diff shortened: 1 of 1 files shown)


bdovaz:
@nvuillam I don't know if anything else needs to be done to make this new field "used" because it actually looks like it was already in use?

Example:

https://github.com/oxsecurity/megalinter/blob/v6.15.0/megalinter/reporters/ConsoleReporter.py#L39

Even though it is not in the schema file it looks like it was previously supported at least by ConsoleReporter?

nvuillam:
@bdovaz instance of linters contain attributes of descriptor level + linter level
What was not implemented is the capability to override it at linter for schema , and you just did that with this PR :)

nvuillam (approved):

PR #2168, opened 2022-12-22
Change reporter text for better UX

ashokm:
To prevent users seeing 'See errors details in' text even when their run is not in an error state, lets change this to 'See detailed report in'.

Also removing the 'artifact' text in the message as there are use cases where artifacts are not published and the link takes the user to the build (and not artifacts) page.

### DIFF
--- a/megalinter/reporters/GitlabCommentReporter.py
+++ b/megalinter/reporters/GitlabCommentReporter.py
@@ -131,8 +131,8 @@ class GitlabCommentReporter(Reporter):
             # Check if there is already a MegaLinter comment
             for comment in existing_comments:
                 if (
-                    "See errors details in [artifact MegaLinter reports on"
-                    in comment.body
+                    "See detailed report in [MegaLinter reports" in comment.body
+                    or "See detailed report in MegaLinter reports" in comment.body
                 ):
                     existing_comment = comment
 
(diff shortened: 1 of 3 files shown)


echoix (approved):
This looks good to me, and a good idea too! Are the other reporters outputting a similar text that should be harmonized too?

nvuillam (changes requested):
Good suggestion, but you forgot a reference in GitlabCommentReporter , please can you update it too ?
image

megalinter/reporters/GitlabCommentReporter.py
```diff
@@ -131,7 +131,7 @@ def produce_report(self):
             # Check if there is already a MegaLinter comment
             for comment in existing_comments:
                 if (
-                    "See errors details in [artifact MegaLinter reports on"
+                    "See detailed report in [MegaLinter reports"
                     in comment.body
```
nvuillam, line 135: Plz can u also test `See detailed report in MegaLinter reports` , as both can happen ? ( add a `or` condition )

nvuillam (changes requested):
plz check last comment :)

nvuillam (changes requested):
plz check comment ^^

nvuillam:
Thanks :)

PR #2175, opened 2022-12-22
PowershellLinter autofix

bdovaz:
- Provide autofix (I have based myself on the example of https://github.com/oxsecurity/megalinter/blob/v6.15.0/megalinter/linters/RakuLinter.py#L38)


bdovaz:
I have the doubt if a test should be created for -Fix? I mean, pass an input and compare an output with the applied fix.

If so, is there any similar example of a linter that does autofix and has tests for this?

@nvuillam I have marked it as draft because the formatter is missing.

nvuillam:
@bdovaz MegaLinter is not perfectly equipped to auto-test formatting and fixes, there is a single test class here, that tests some linter fixes but not all of them : https://github.com/oxsecurity/megalinter/blob/main/megalinter/tests/test_megalinter/mega_linter_2_fixes_test.py

About formatter, as it may be longer to build, you may do it in a separate PR ^^

bdovaz:
@nvuillam Ready!

nvuillam (approved):
Great :)

Issue #2202, opened 2022-12-30
Improve the documentation of cli_lint_mode


bdovaz:
@nvuillam There are linters that support more than one mode such as:

https://github.com/oxsecurity/megalinter/blob/v6.17.0/megalinter/linters/JavaPmdLinter.py#L38

On the other hand, the "cli_lint_mode" field only supports one value and not N:

https://github.com/oxsecurity/megalinter/blob/v6.17.0/megalinter/descriptors/java.megalinter-descriptor.yml#L76

"cli_lint_mode" should be an array (and rename it to "cli_lint_modes") and add the modes supported (apart from another "cli_lint_mode_default" or similar)?

```yaml
cli_lint_modes:
  - file
  - list_of_files
  - project
cli_lint_mode_default: list_of_files
```

In the code that generates the documentation it assumes that if it is not of type "project" it will support everything and it does not have to:

https://github.com/oxsecurity/megalinter/blob/v6.17.0/.automation/build.py#L1246

The documentation should be generated based on what is in the descriptor. Right now it is misleading.

If you set "*_CLI_LINT_MODE" in a linter to a value that is not supported by the linter, what happens? Is there any code that controls this?

nvuillam:
@bdovaz there is already a doc page about it :) https://megalinter.io/latest/configuration/#cli-lint-mode

It is not something usual to override cli_lint_mode, this mode is mainly for internal use :)

If a linter is defined with cli_lint_mode: project, it can not be overriden to file or list_of_files, or to from file to list_of_files -> https://github.com/oxsecurity/megalinter/blob/d2b1417774be5df5b9b263e7b80d72337d62fcd3/megalinter/Linter.py#L564

bdovaz:
@nvuillam looking at that documentation (https://megalinter.io/latest/configuration/#cli-lint-mode) I see that:
1. if "cli_lint_mode: project" can be changed neither to file nor to list_of_files.
2. It does not explain that if "cli_lint_mode: file" can not be changed neither to list_of_files nor to project (in fact, this last one is not controlled in the code that you have put me, I understand that it is a bug...).

For example: https://megalinter.io/latest/descriptors/bicep_bicep_linter/#readme

It has "cli_lint_mode: file" in its descriptor but according to that documentation in the field "BICEP_BICEP_LINTER_CLI_LINT_MODE" it says that you can put any of those 3 values when it is not true.

[...]

PR #1877, opened 2022-09-15
Use relative file paths to call linters

nvuillam:
Fixes https://github.com/oxsecurity/megalinter/issues/1875

PR #2206, opened 2022-12-31
Lint mode docs improvements

bdovaz:
Context https://github.com/oxsecurity/megalinter/issues/2202

@nvuillam I have doubts if it makes sense that this row and the schema enum appears if "cli_lint_mode: file" because in this case there is no other possible value. It would be the equivalent case to when "cli_lint_mode: project" then neither the row nor the schema is generated.

Check everything in case there is a case that I have not understood correctly.


nvuillam (commented):
If cli lint mode == project, only available value in enum should be project, is this taken in account ?

nvuillam:
And json schema must be generated again with bash build.sh :)

nvuillam (changes requested):
Plz check comments :)

bdovaz:
> If cli lint mode == project, only available value in enum should be project, is this taken in account ?

@nvuillam Right now in the original code if it is project that entry does not appear in the table not even: https://github.com/oxsecurity/megalinter/blob/main/.automation/build.py#L1257

It is necessary to decide if in the case of == file as well as in the case of == project if that entry should appear or not with the only value of each case or not. Currently as I say if == project, it does not appear.

nvuillam:
Agreed, if there is only one value possible, no need to make it appear in the variables list in the doc :)

bdovaz:
> And json schema must be generated again with bash build.sh :)

@nvuillam Do you mean "bash build.sh --doc"? Because the only thing I change with this code is documentation nothing else.

If I run "bash build.sh" nothing changes.

nvuillam:
That's strange, add_in_config_schema_file should update the json schema to remove obsolete enum values...

bdovaz:
> That's strange, add_in_config_schema_file should update the json schema to remove obsolete enum values...

@nvuillam But that's what I'm telling you, the code involving what I've already discussed:

https://github.com/oxsecurity/megalinter/blob/main/.automation/build.py#L1257

It is invoked only by adding --doc by generate_documentation() method

nvuillam:
and even when you call bash build.sh --doc it does not update the JSON schema ?

bdovaz:
> and even when you call bash build.sh --doc it does not update the JSON schema ?

@nvuillam Yes, of course in that case yes but originally you had told me to run "bash build.sh" without "--doc" hence the confusion in the conversation I think.

nvuillam:
You acn run with the `--doc` in special cases like this (it just updates more files so create more conflicts, that's why the doc says to run only `bash build.sh` by default :) )

bdovaz:
@nvuillam I have one last doubt.

We have made it clear that neither "cli_lint_mode: file" nor "cli_lint_mode: project" allow to override the *_CLI_LINT_MODE variable and that is why it should not appear in the documentation.

You also said that if "cli_lint_mode: list_of_files" is implicit to allow linting in "cli_lint_mode: file" mode, but what happens in that case with the "project" mode? All linters that have "cli_lint_mode: list_of_files" do not necessarily have to work in "project" mode, right?

nvuillam:
list_of_files is the most optimized way to run linters (because identifying files has already been done by MegaLinter), contrary to file which is performance hell (but can be required for some rare cases like TAP output). For those few cases, user must be able to go from list_of_files to file
default project mode is mostly for linters that can not accept file or list_of_file, like security linters, that's why it's not possible to override the value
Allowing file or list_of_files to be overridden to project is mostly for workarounds, for example with linters that have a problem to find their config file when the current folder is not the repo root (that will be solved in the infinite PR #1877 )

megalinter/Linter.py
```diff
@@ -571,10 +571,10 @@ def load_config_vars(self, params):
                 )
             elif (
                 cli_lint_mode_descriptor == "file"
-                and cli_lint_mode_config == "list_of_files"
+                and (cli_lint_mode_config == "list_of_files" or cli_lint_mode_config == "project")
```
nvuillam, line 574: nah, project can work as override of file mode ^^

bdovaz: Done

bdovaz:
@nvuillam I think it's ready!

nvuillam:
this CI error arrives from time to time, it's ok to bypass it

nvuillam (approved):

PR #2220, opened 2023-01-04
Unify the drawing of badges

bdovaz:
- Badges now are generated in all places where the linter is mentioned (descriptor, individual descriptor and flavor)
- Unify the generation of badges in a single function: **get_badges(linter)**
- Add deprecated linter badge to **get_badges(linter)**
- Add missing **linter_repo** fields: In the **get_repository_badge_url(linter)** method it takes into account and tries to search in the **linter_url** field if the **linter_repo** field does not exist but as it is not a unified logic everywhere, I have added the fields in the missing linters


nvuillam:
That seems to be a nice refactoring PR :)

But I (may) remember that for some doc pages, having less badges was on purpose to avoid ugly displays in tables

Please can you try locally with `mkdocs serve` and check that the doc display is still correct  ? 

```shell
pip install --upgrade "markdown==3.3.7" mike mkdocs-material mdx_truly_sane_lists jsonschema json-schema-for-humans giturlparse webpreview github-dependents-info
mkdocs serve
```

nvuillam:
The last commit is not so relevant... especially not in first position ... maybe just remove it ?

bdovaz:
> The last commit is not so relevant... especially not in first position ... maybe just remove it ?

@nvuillam do you think so? The color of that particular badge gives at a glance the health of that linter in terms of maintenance. It is not feasible to periodically monitor the status of each linter one by one.

The information on that badge tells you things like if it is time to mark a linter as deprecated.

To appear on pages like this: https://megalinter.io/latest/supported-linters/ allows at a glance to visualize the health status of each linter in a very simple way.

If it is true that the fact that it appears first of all is not logical. What do we do then? 
1. We move it to the last position + the change I have made to shorten the name?
2. We decide page by page (descriptor, individual descriptor, flavor...) whether to show it or not depending on how it is displayed (under the title or in table)?

I opt for the first one but well, I am open to other opinions.

nvuillam:
I'm more team solution 2, because on the linter doc page it's great to have lots of badges (we could even add more !), but in tables we need a restricted one (stars + formatter + SARIF ?)

For linter doc page, we could have: 
- same than table badges +
- Github last release: GitHub release
- Github last commit: [GitHub last commit](https://github.com/oxsecurity/megalinter/commits/)
- Github commits activity: [GitHub commit activity](https://github.com/oxsecurity/megalinter/graphs/commit-activity/)
- Github contributors: [GitHub contributors](https://github.com/oxsecurity/megalinter/graphs/contributors/)

nvuillam:
While you're at it, if you can remove the column configuration key from https://megalinter.io/latest/supported-linters/ , it will make a better display ^^

nvuillam:
(horrible horizontal scrollbar in the table here :/ )

image

bdovaz:
> While you're at it, if you can remove the column configuration key from https://megalinter.io/latest/supported-linters/ , it will make a better display ^^

@nvuillam I'm trying to do that but when I remove the "Configuration key" column from each page of a flavor, I get all the rows in the table disappearing.

I think it has to do with `<!-- linter-icon -->` because if I remove that it doesn't happen anymore. Can it be that when deleting a column, some calculation is made with some row that makes disappear what it should not? I know that in those pages there is some kind of logic to hide repeated icons, isn't it? Example: https://megalinter.io/latest/flavors/documentation/

nvuillam:
If you remove from the table header + the element in the libe it should be ok :)

But don't worry I can do that in a PR after yours:)

bdovaz:
@nvuillam I think that with this commit you would have everything you asked for, except for the flavors and `<!-- linter-icon -->`.

nvuillam:
@bdovaz  sorry I validated the dotnet v6 PR, lots of conflicts ^^

bdovaz:
@nvuillam conflicts resolved!

nvuillam (approved):
Great PR R by a great contrbutor, again :)

bdovaz:
@nvuillam the problem of flavors tables being empty is still a problem, did you finally look at it?

Example:

https://megalinter.io/beta/flavors/dotnet/

nvuillam:
woooo didn't see this one :o

nvuillam:
@bdovaz found it :)

PR #2235, opened 2023-01-08
Fix broken link

bdovaz:
I didn't realize that in the documentation you can't put links relative to the repo, I change it to absolute.


nvuillam (approved):
Because of online html generated from /docs markdown files, you can use relative links if they are targeting something in /docs folder, else indeed it needs to be absolute:)

PR #2236, opened 2023-01-08
Fix local files with extends

bdovaz:
This PR #2151 that I made was not right and we didn't realize it because we didn't have tests.

In this PR I add the fix and the tests

@nvuillam check me the tests in case you see that they can be done better because my experience with python is not much and I don't know if the way to build the workspace value is the best.


nvuillam:
tips here about quick build and TEST_KEYWORDS, that can dramatically improve CI jobs time when you update just python or doc :)

https://megalinter.io/latest/contributing/#cictcd

image

About the last bullet, for a PR like that I'll merge without requiring it :)

bdovaz:
Thanks @nvuillam

By the way, checks validated!

nvuillam (approved):
Good catch and good fix :)

Issue #2240, opened 2023-01-09
Support --format with xmllint


bdovaz:
@nvuillam has not raised this issue so far?

It looks like you can at least configure the indent with the `XMLLINT_INDENT` environment variable:

https://gnome.pages.gitlab.gnome.org/libxml2/xmllint.html

nvuillam:
- I didn't know it was possible ... but i didn't check, xmllint is inherited from super-linter who can not apply updates

Maybe we should allow it but disable it by default?

bdovaz:
@nvuillam Questions:
1. How do I disable it by default? In the descriptor I have added `cli_lint_fix_arg_name: "--format"`.
2. The environment variable `XMLLINT_INDENT` (string) is the one that controls the number of spaces:
    * How can I configure the linter so that this setting can be passed to it like any other linter configuration?
    * A class is needed for this, isn't it? Like a XmlLintLinter.py in order to redirect this configuration to this process

nvuillam:
1. The current architecture does not allow to do that :D We have to think about it... maybe an additional variable XML_XMLLINT_AUTOFORMAT , that could be used in XmlLintLinter.py

2 descriptors allow to define custom variables :)
They will appear in the doc after build.sh --doc, but I'm not sure if it also updates JSON schema (in that case it must be done manually)

Example: https://github.com/oxsecurity/megalinter/blob/c5561aa7a644977620f16ee6d30278513adab1e6/megalinter/descriptors/spell.megalinter-descriptor.yml#L38

PR #2244, opened 2023-01-10
Support xmllint autofix

bdovaz:
Context #2240 

It will only work if `XML_XMLLINT_CLI_LINT_MODE: file` because the command needs an `--output file` which can only be passed per file.

I also created 2 files for the tests because the linter was of type `list_of_files` but only had one file of each.

@nvuillam I have doubts in this case and in the powershell one and it is that in all linter tests the autofix are never tested, no? I mean, there should be a test that checks the original content of a file and the expected one to see if it has been formatted correctly. Right now without those tests, we don't know if it really works or not.


megalinter/linters/XmlLintLinter.py
```diff
@@ -0,0 +1,29 @@
...
+    def build_lint_command(self, file=None):
+        cmd = super().build_lint_command(file)
+
+        if (
+            self.apply_fixes is True
+            and self.cli_lint_fix_arg_name is not None
+            and config.get("XML_XMLLINT_AUTOFORMAT", "false") == "true"
+        ):
+            if self.cli_lint_mode == "file":
+                os.environ["XMLLINT_INDENT"] = config.get("XML_XMLLINT_INDENT", "  ")
```
bdovaz, line 22: @nvuillam this is correct, isn't it? As I already told you, the fact of how many spaces it applies is done through an environment variable

megalinter/descriptors/xml.megalinter-descriptor.yml
```diff
@@ -8,15 +8,25 @@ file_extensions:
...
     linter_url: http://xmlsoft.org/xmllint.html
     linter_rules_url: http://xmlsoft.org/xmllint.html#diagnostics
     version_extract_regex: "(?<=libxml version )\\d+(\\d+)+"
     cli_lint_mode: list_of_files
+    cli_lint_fix_arg_name: "--format"
     cli_help_arg_name: "--help"
     help_command_return_code: 1
     examples:
       - "xmllint myfile.xml"
+      - "xmllint --format myfile.xml --output myfile.xml"
```
nvuillam, line 29: this example is not existing, no ?
maybe better to write the real examples like `XML_XMLLINT_AUTOFORMAT xmllint myfile.xml myfile2` ?

megalinter/linters/XmlLintLinter.py
nvuillam: why defining indent only if cli_lint_mode = file ? not for all modes (including list_of_files) ?

nvuillam: oh ok I see the code ^^ autoformat works only with one file as parameter ??

megalinter/descriptors/xml.megalinter-descriptor.yml
```diff
@@ -8,15 +8,25 @@ file_extensions:
...
+        default_value: "false"
+        description: If set to `true`, it will reformat and reindent the output
+      - name: XML_XMLLINT_INDENT
+        default_value: "  "
+        description: The number of indentation spaces when `XML_XMLLINT_AUTOFORMAT` is `true`
     linter_url: http://xmlsoft.org/xmllint.html
     linter_rules_url: http://xmlsoft.org/xmllint.html#diagnostics
     version_extract_regex: "(?<=libxml version )\\d+(\\d+)+"
     cli_lint_mode: list_of_files
+    cli_lint_fix_arg_name: "--format"
```
nvuillam, line 24: When this is not the real argument string, use the same dummy one than for others so it won't be added in the command :)
  `  cli_lint_fix_arg_name: "--megalinter-fix-flag" # Workaround for MegaLinter `

megalinter/descriptors/xml.megalinter-descriptor.yml
```diff
@@ -8,15 +8,25 @@ file_extensions:
   - ".xml"
 linters:
   # XML LINT
-  - linter_name: xmllint
+  - class: XmlLintLinter
+    linter_name: xmllint
+    variables:
```
nvuillam, line 13: Probably add some comments in linter_description so it will apear at the top of the doc (people won't think to read all the variables )

nvuillam (changes requested):
plz check comments :)

nvuillam:
https://github.com/oxsecurity/megalinter/blob/main/megalinter/tests/test_megalinter/mega_linter_2_fixes_test.py

You can test the fix here :)

megalinter/descriptors/xml.megalinter-descriptor.yml
bdovaz: What do you mean it does not exist? I have based on the first example that was already there and looking at other descriptors, they also add similar examples.

I don't understand the example you give me either, that line is not a valid command.

megalinter/descriptors/xml.megalinter-descriptor.yml
bdovaz: As you have already deduced, the `--format` can only be used in "file" mode because as I have put in the first message of this PR, it expects an `--output` parameter which evidently has to be file to file.

Why do you say that `--format` is not the actual argument? It is:

image

image

megalinter/descriptors/xml.megalinter-descriptor.yml
bdovaz: Done

bdovaz:
> https://github.com/oxsecurity/megalinter/blob/main/megalinter/tests/test_megalinter/mega_linter_2_fixes_test.py
> 
> You can test the fix here :)

@nvuillam But what exactly do you mean?

I see that `test_1_apply_fixes_on_one_linter` is a JavaScript only test.

I see that `test_2_apply_fixes_on_all_linters` is a test that mixes everything, JavaScript, several different files in `fixable_files`...

Where do I have to add what? They are a bit messy those tests.

nvuillam:
Yep it's messy but except you want to implement generic testing of fixing (that is an option but would require core enhancements & probably more test files), this classe is the only thing testing fixes today :/

bdovaz:
@nvuillam please check comment replies...

About the tests you mention... For them to run correctly locally (on my machine) I have to have all the linters installed on my machine? I see that some of them work and others fail and I understand that is why, isn't it? How can I see the complete log? Because in VSCode I get the output but I see it like this, is it normal?

image

image

nvuillam:
Hmmm indeed, run this test class locally can be tricky if all the linters locally installed are not running the same on windows & linux...

If your branch was directly on the repo and not from a fork, CI generates and push the docker image named like the branch, so it can be used for tests on another sample repo

Probably someday we should implement a generic way to test fixes are applied, like we do for success, failure, version, help and sarif... (but i'm afraid it makes the CI jobs even more longer :/ )

megalinter/descriptors/xml.megalinter-descriptor.yml
nvuillam: ok i didn't know it was, i thought it was managed with another env var

megalinter/descriptors/xml.megalinter-descriptor.yml
nvuillam: oops, i didn't understand , my bad ^^

nvuillam (approved):
ok once CI is ok :)

Kurt-von-Laven:
[Hypothesis](https://hypothesis.readthedocs.io/en/latest/) is a phenomenal property-based testing library for Python, but we're a ways off from that today.

echoix:
> [Hypothesis](https://hypothesis.readthedocs.io/en/latest/) is a phenomenal property-based testing library for Python, but we're a ways off from that today.

But we are testing that our wrapping code and calling the linter works, not testing the linter itself. As everyone, learning to better QA software is always a good thing. It's always a weak spot.

Kurt-von-Laven:
Hypothesis can help with that too, but it's rarely the case that one would want to write all tests exclusively in Hypothesis. Property-based testing is intended to be used in conjunction with conventional example-based testing.

bdovaz:
We take the testing conversation to #2250

If the PR is ok @nvuillam you can merge it

Issue #2300, opened 2023-01-30
Sarif Reporter in Azure Devops with space in project name


EtienneDeneuve:
**Describe the bug**
Report in Azure Devops with Sarif reporter enabled can have broken link when the project have a whitespace in the name.

**Expected Behaviour**
Replace the whitespace with url encoded string '%20'

2022H1030042G:
Hi, I would like to work on this issue, is it available?

EtienneDeneuve:
Normally, I fixed it already, but you can test
________________________________

EtienneDeneuve:
If the image is build, for sure, I'll do it in the day

nvuillam:
@EtienneDeneuve it is, ghcr.io is back for a while :)

PR #2301, opened 2023-01-30
WIP: Change the way Azure Devops project 

EtienneDeneuve:
Should fix issue #2300

### DIFF
--- a/megalinter/reporters/AzureCommentReporter.py
+++ b/megalinter/reporters/AzureCommentReporter.py
@@ -12,7 +12,7 @@ Requires the following vars sent to docker run:
 - BUILD_REPOSITORY_ID
 """
 import logging
-
+import urllib.parse
 import requests
 from megalinter import Reporter, config
 from megalinter.utils_reporter import build_markdown_summary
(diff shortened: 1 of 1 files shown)


Kurt-von-Laven (commented):
Should we do the same with `SYSYSTEM_COLLECTIONURI`?

EtienneDeneuve:
I was searching, I've a doubt if we can create an Azure DevOps org with a white space.

Many company use space in the project or repo name, maybe it's better to sanatize all, just in case, don't you think so?
________________________________

nvuillam:
By definition, an URI does not need to be encoded / decoded , so if SYSTEM_COLLECTIONURI is a valid URI, it should not be necessary to manipulate it :)

nvuillam (approved):
Thanks for the catch and thanks for the PR :)

nvuillam:
The megalinter-runner job issue is not related to this PR, that's why I force its merge

EtienneDeneuve:
And thanks for megalinter and the merge ;)
________________________________

nvuillam:
@EtienneDeneuve you're welcome :)
Posting on Azure is not in the GitHub related CI... if you can confirm that your fix well with docker image **oxsecurity/megalinter:beta** before our next release (that should happen soon) , it would be nice ;)

Note: ghcr.io seems down, or in bad shape... so the build of beta is currently failing ... when it will be available I'll notify here :)

PR #2320, opened 2023-02-06
WIP: Add support for job arrays based on worker index/total variables

### LINKED ISSUE #2303: Concurrent parallel jobs in Gitlab CI (and others)
We can't currently make full use of available resources on other machines

An option to only run a subset of selected linters based on environment variables that indicate the total number of jobs in the array and the current job's index in the array using CI features like [Gitlab's `parallel:` directive](https://docs.gitlab.com/ee/ci/jobs/job_control.html#parallelize-large-jobs)

Hacked together the following solution myself, but this could be much more efficient properly integrated:
```python
 from itertools import cycle
 from megalinter import Megalinter, linter_factory

 megalinter = Megalinter({"cli": True})
 # deactivates linters that dont match any files
 megalinter.collect_files()

 active_linters = sorted((linter for linter in megalinter.linters if linter.is_active), key=lambda x: x.linter_speed)

 node_index = int(os.environ.get("CI_NODE_INDEX")) - 1 #Gitlab env var isn't zero indexed
 node_total = int(os.environ.get("CI_NODE_TOTAL"))

 # create an empty list for each node in the job array
 node_buckets = [[] for node in range(node_total)]
 
# round robin through the buckets, dropping linters in each bucket in order of linter_speed
 queue = cycle(node_buckets)
 for linter in active_linters:
    	next(queue).append(linter)

 with open('job_linters', 'w') as f:
   f.write(','.join(linter.name for linter in node_buckets[node_index]))
```
The initialization of the class outputs a bunch of stuff that doesn't seem to have a way to disable, so I write the list of linters to a file and then export the content of that file as `ENABLE_LINTERS`.

I've also made use of the linter_speed attribute in a naive attempt to normalize the load across all jobs in the array.

bstrdsmkr:
Fixes #2303 

## Proposed Changes

Allow for the dynamic splitting of linters over a number of workers based on a pair of variables -- either:
  1. New `MEGALINTER_ARRAY_INDEX` and `MEGALINTER_ARRAY_SIZE`
      -   These take precedence over the variables provided by common CI/Job scheduling platforms
  1. [Gitlab CI's `CI_NODE_INDEX` and `CI_NODE_TOTAL`](https://docs.gitlab.com/ee/ci/yaml/#parallel)
  1. [Circle CI's `CIRCLE_NODE_INDEX` and `CIRCLE_NODE_INDEX`](https://circleci.com/docs/parallelism-faster-jobs/#using-environment-variables-to-split-tests)
  1. [SLURM's `SLURM_ARRAY_TASK_ID` and `SLURM_ARRAY_TASK_ID`](https://slurm.schedmd.com/job_array.html#env_vars)

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/megalinter/tests/test_megalinter/job_arrays_test.py
+++ b/megalinter/tests/test_megalinter/job_arrays_test.py
@@ -1,0 +1,88 @@
+"""
+Unit tests for Megalinter class
+
+"""
+import os
+import unittest
+
+import megalinter
+from megalinter.tests.test_megalinter.helpers import utilstest
+from unittest.mock import create_autospec
+from unittest.mock import patch
+
+
+class job_array_configuration_test(unittest.TestCase):
+
+    def test_job_array_vars_are_parsed(self):
+        """
+        Test that the various job array variables are parsed correctly
+        """
+
+        scenario = [
+            # Style                           Inputs                                       Expected
+            ("Native",    {"MEGALINTER_ARRAY_INDEX": "5", "MEGALINTER_ARRAY_SIZE":  "10"}, (5, 10)),  # nopep8
+            ("Gitlab",    {"CI_NODE_INDEX":          "5", "CI_NODE_TOTAL":          "10"}, (4, 10)),  # nopep8
+            ("Circle CI", {"CIRCLE_NODE_INDEX":      "5", "CIRCLE_NODE_TOTAL":      "10"}, (5, 10)),  # nopep8
+            ("SLURM",     {"SLURM_ARRAY_TASK_ID":    "5", "SLURM_ARRAY_TASK_COUNT": "10"}, (5, 10)),  # nopep8
+        ]
+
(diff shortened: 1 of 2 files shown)


megalinter/tests/test_megalinter/job_arrays_test.py
```diff
@@ -0,0 +1,88 @@
...
+
+class job_array_configuration_test(unittest.TestCase):
+
+    def test_job_array_vars_are_parsed(self):
+        """
+        Test that the various job array variables are parsed correctly
+        """
+
+        scenario = [
+            # Style                           Inputs                                       Expected
```
Kurt-von-Laven, line 22: "Style" -->" "Provider" to match the variable name.

Kurt-von-Laven (changes requested):
It seems like a good start to me. I like the Pytest-inspired approach used in the testing; it saves a lot of duplication.

PR #2327, opened 2023-02-08
Fix pre-commit CLI parameter syntax & sort parameters

sanmai-NL:
As is, optionator rejects the command line parameter and argument syntax.

```
Run MegaLinter...........................................................Failed
- hook id: megalinter-full
- exit code: 1

/Users/user/.npm/_npx/614e542633a698fb/node_modules/optionator/lib/index.js:316
          throw new Error("Value for '" + prop + "' of type '" + getOption(prop).type + "' required.");
                ^

Error: Value for 'container-name' of type 'String' required.
...
```

## Proposed Changes

1. Fix pre-commit parameter syntax.

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -5,10 +5,7 @@
   require_serial: true
   args:
     - mega-linter-runner
-    - --containername
-    - "megalinter-incremental"
-    - --remove-container
-    - --fix
+    - --containername=megalinter-incremental
     - --env
     - "'APPLY_FIXES=all'"
     - --env
@@ -16,6 +13,8 @@
     - --env
     - "'LOG_LEVEL=warning'"
     - --filesonly
+    - --fix
+    - --remove-container
   description: >
     See https://megalinter.io/latest/mega-linter-runner/#usage
     and https://megalinter.io/latest/configuration/ if you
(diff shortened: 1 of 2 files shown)


nvuillam:
@Kurt-von-Laven do you have an opinion about this PR ? (i'm not a pre-commit expert ^^ )

haydenseitz:
Not sure if code was already merged to address this, but updating from 6.8.0 to 6.19.0 doesn't have the error for me when using pre-commit.

nvuillam:
The great @Kurt-von-Laven will decide about this topic :)

Kurt-von-Laven:
@sanmai-NL, can you provide reproduction steps? We have been using one variant or another of the pre-commit hooks before they were even added to MegaLinter and have never encountered this error. MegaLinter doesn't use optionator, so I would need more context to understand the rationale for this pull request. What are you trying to accomplish that motivated this change?

nvuillam:
@Kurt-von-Laven mega-linter-runner uses Optionator ^^
https://github.com/oxsecurity/megalinter/blob/main/mega-linter-runner/lib/options.js

sanmai-NL:
> @sanmai-NL, can you provide reproduction steps? We have been using one variant or another of the pre-commit hooks before they were even added to MegaLinter and have never encountered this error. MegaLinter doesn't use Optionator, so I would need more context to understand the rationale for this pull request. What are you trying to accomplish that motivated this change?

I'm using this hook config directly in my pre-commit config. Then the fault pops up, unless my fix is applied. 
this happens with a straight:

```sh
pre-commit run --all-files megalinter-full
```

I'm using pre-commit 3.0.4.

Kurt-von-Laven:
I am on Linux rather than macOS but the same version of pre-commit and am unable to reproduce this issue at MegaLinter v6.19.0 and mega-linter-runner v6.19.0. What version of MegaLinter and mega-linter-runner are you on? Maybe @haydenseitz was onto something about the issue already having been fixed?

sanmai-NL:
Not stale.

Kurt-von-Laven:
I would rebase at this point since I suspect your issue has already been addressed in v6.20.0.

sanmai-NL:
I haven't had time to investigate further. But I haven't read anything here that convinces me this is fixed.

Kurt-von-Laven:
#2411

PR #2383, opened 2023-02-21
feat: add option to skip def_ws prefix in sarif reports

### LINKED ISSUE #2006: DEFAULT_WORKSPACE is prefixed to SARIF-output field uri but should not be.

janderssonse:
Fixes #2006 
 
This PR is a suggestion for solving the use case
of needing to remove the DEFAULT_WORKSPACE from
the out put in the generated SARIF output.
(https://github.com/oxsecurity/megalinter/issues/2006).

It moves the SARIF logic to an earlier phase, to be handled before the aggregate SARIF generation.
It replaces the prefix if the flag
SARIF_REPORTER_NORMALIZE_LINTERS_OUTPUT: true is set (default: true)
(see clear_default_workspace_prefix(...

Implementation is done by line parsing and replacing, as a node traversal solution quickly would grew due to
the many places in the sarif out put the uri can be found (metris, relatedLocations, and so on), and the code is much simpler this way to maintain.

Improvements and suggestions:
- Could json dumps() and resulting json string be used in
a reliable way to stream line parse an json file? I didn't find a good way.
- Should the option be renamed to
SARIF_REPORTER_DISABLE_DEFAULT_WORKSPACE_IN_OUTPUT or alike. As the pre existing normalization still happens? (We don't change that pre existing behaviour in this PR, only the DEFAULT_WORKSPACE prefix part).

## Proposed Changes

1. Move the SARIF Logic to an earlier phase 
2. Add optional removal of DEFAULT_WORKSPACE prefix
3. Activated by SARIF_REPORTER_NORMALIZE_LINTERS_OUTPUT

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/docs/reporters/SarifReporter.md
+++ b/docs/reporters/SarifReporter.md
@@ -2,6 +2,7 @@
 title: SARIF Reporter for MegaLinter
 description: Generates SAST results in SARIF format within a file named mega-linter-report.sarif, located in report folder
 ---
+
 # SARIF Reporter (beta)
 
 Generates a full execution log in SARIF format within a file named **mega-linter-report.sarif** , located in report folder.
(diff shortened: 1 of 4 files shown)


nvuillam:
That seems great :)

I think SARIF_REPORTER_NORMALIZE_LINTERS_OUTPUT  should be true by default :)

janderssonse:
Thanks, fixed so it is default true now. Btw,How can I make the checks failing tests succed, if i interprets the checks logs correct they would need permission conf (PAT)? Learning for this and future fixes.:)

Kurt-von-Laven:
I believe the errors are the linter violations, not the bit about the PAT, which you can ignore.

janderssonse:
> I believe the errors are the linter violations, not the bit about the PAT, which you can ignore.
Uh, oh, I see now, preparing a fix for these.

nvuillam (approved):
Indeed green, well done , thanks :):

PR #2413, opened 2023-03-04
Fix banner height

nvuillam:


echoix:
Oh thanks! That was so much bothering me on Monday (old screenshot)
image

nvuillam:
Saw that on my phone... i was horrified :D

PR #2440, opened 2023-03-07
Correct minor docs error

moverperfect:
## Proposed Changes

1. This sentence should suggest to the user that they should replace main with master in the URLs not main with main

### Author/Contributor

### Reviewing Maintainer


echoix (approved):

nvuillam (approved):
It seems that we made too many replacements when migrating from master to main, thanks for catching :)

PR #2510, opened 2023-04-01
AzureCommentReporter vars reflect official Azure DevOps naming

### LINKED ISSUE #2509: (AzureCommentReporter) BUILD_BUILD_ID is not official Azure DevOps env var
## Describe the bug

BUILD_BUILD_ID is not official Azure DevOps env var (ref: https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#build-variables-devops-services) used in AzureCommentReporter

In the AzureCommentReporter + docs all env vars follows official naming except this one.

## To Reproduce

1. Run megalinter on Azure DevOps with AzureCommentReporter using official build agent env vars
2. See the link under the PR comment: `See detailed report in [MegaLinter reports]`
3. Link in not valid and points to `...results?buildId=None&view=artifacts&pathAsName=false&type=publishedArtifacts` (`buildId` is `None`)

## Expected behavior

Link should point to valid buildId when official Azure DevOps env vars are used -> `BUILD_BUILDID`

DariuszPorowski:
Fixes #2509 

## Proposed Changes

1. Changed vars naming in AzureCommentReporter vars to reflects official Azure DevOps naming convention
2. Updated docs with Azure Pipelines usage scenario with passing build agent env vars - it eliminates explicite passing vars one-by-one (except SYSTEM_ACCESSTOKEN - it requires to be pass explicitly) and gives opportunity to work with ADO agent vars in the future. 
   - the rest env vars can be pure megalinter envs, it increases cleanness in the pipeline
4. Using old BUILD_BUILD_ID as fallback to keep backward compatibility
5. Not planned, but came up while working on it this fix: updated outdated ToC in main README + minor linting

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/README.md
+++ b/README.md
@@ -531,6 +531,7 @@ Add the following job in your `azure-pipelines.yaml` file
       - script: |
           docker run -v $(System.DefaultWorkingDirectory):/tmp/lint \
             --env-file <(env | grep -e SYSTEM_ -e BUILD_ -e TF_ -e AGENT_) \
+            -e CI=true \
             -e SYSTEM_ACCESSTOKEN=$(System.AccessToken) \
             -e GIT_AUTHORIZATION_BEARER=$(System.AccessToken) \
             oxsecurity/megalinter:v6
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ _Github PR reporter_
 <!-- table-of-contents-start -->
 ## Table of Contents
 
-- [MegaLinter, by](#megalinter-by-)
+- [MegaLinter](#megalinter-by-)
   - [Table of Contents](#table-of-contents)
   - [Why MegaLinter](#why-megalinter)
   - [Quick Start](#quick-start)
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 <!-- header-logo-end -->
 <!-- mega-linter-title-start -->
 
-## MegaLinter, by [![OX Security](https://www.ox.security/wp-content/uploads/2022/06/logo.svg?ref=megalinter_readme)](https://www.ox.security/?ref=megalinter)
+# MegaLinter, by [![OX Security](https://www.ox.security/wp-content/uploads/2022/06/logo.svg?ref=megalinter_readme)](https://www.ox.security/?ref=megalinter)
 
 <!-- mega-linter-title-end -->
 <!-- mega-linter-badges-start -->
@@ -73,7 +73,7 @@ _Github PR reporter_
 <!-- table-of-contents-start -->
 ## Table of Contents
 
-- [MegaLinter](#megalinter)
+- [MegaLinter, by](#megalinter-by-)
   - [Table of Contents](#table-of-contents)
   - [Why MegaLinter](#why-megalinter)
   - [Quick Start](#quick-start)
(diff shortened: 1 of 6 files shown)


nvuillam:
Great initiative that will reduce the size of the Azure pipeline script, many thanks :)

As you seem to know Azure a lot, do you recommend a python package to call Azure API ?

I'd like to enhance azure integration by removing the previous MegaLinter threads on PRs when posting a new one, and calling directly with http does not feel easy to maintain

DariuszPorowski:
hi @nvuillam the only what I know, and its official is https://github.com/microsoft/azure-devops-python-api it's like wrapper-style sdk for rest.

nvuillam:
Thanks, I'll use it in a future PR :)

README.md
```diff
@@ -527,15 +530,9 @@ Add the following job in your `azure-pipelines.yaml` file
       # Run MegaLinter
       - script: |
           docker run -v $(System.DefaultWorkingDirectory):/tmp/lint \
-            -e GIT_AUTHORIZATION_BEARER=$(System.AccessToken) \
-            -e CI=true \
```
nvuillam, line 531: **CI** variable is used within MegaLinter so should be kept in the docker run arguments:)

docs/installation.md
```diff
@@ -192,15 +192,9 @@ Add the following job in your `azure-pipelines.yaml` file
       # Run MegaLinter
       - script: |
           docker run -v $(System.DefaultWorkingDirectory):/tmp/lint \
-            -e GIT_AUTHORIZATION_BEARER=$(System.AccessToken) \
-            -e CI=true \
```
nvuillam, line 196: **CI** variable is used within MegaLinter so should be kept in the docker run arguments:)

nvuillam (changes requested):
Great quality PR, but plz check comments to make a tiny fix before we can merge :)

README.md
DariuszPorowski: @nvuillam hmm, interesting... I tried to figure out purpose of this env, and only found reference here: https://github.com/oxsecurity/megalinter/blob/main/mega-linter-runner/test/megalinter-cli.test.js#L43 what is not related with scanning.
I did all my tests on my Azure DevOps instance without it as well and not observed any issues :) maybe it's obsoleted?

nvuillam: It is used here :)
image

nvuillam (approved):
Great great PR :)

README.md
DariuszPorowski: @nvuillam hah! I tried config.get("CI"), os.environ.CI, etc but not tried simple "CI" 😅 thanks

Regardless of that, I think CI is not necessary here and below code is enough:

```python
if config.get("CONSOLE_REPORTER_SECTIONS", "true") == "true":
```
because the rest part of the code check if tool is ci and not another rule and conditions apply
https://github.com/oxsecurity/megalinter/blob/main/megalinter/utils_reporter.py#L199
https://github.com/oxsecurity/megalinter/blob/main/megalinter/utils_reporter.py#L203
https://github.com/oxsecurity/megalinter/blob/main/megalinter/utils_reporter.py#L207

eventually to keep backward compatibility and get rid of CI env in pipeline's code:

```python
if is_ci() and config.get("CONSOLE_REPORTER_SECTIONS", "true") == "true":
```

```python
def is_ci() -> bool:
    return True if "CI" in os.environ or is_github_actions() or is_gitlab_ci() or is_azure_pipelines() else False
```

nvuillam: Agreed :)
Can i let you do that in https://github.com/oxsecurity/megalinter/pull/2504 ? :)

PR #2619, opened 2023-05-07
Remove additional `--update` for apk in Dockerfile

PeterDaveHello:
There is no need to use `--update` with `--no-cache` when using `apk` on Alpine Linux, as using `--no-cache` will fetch the index every time and leave no local cache, so the index will always be the latest without temporary files remain in the image.

## Proposed Changes

- Remove `--no-cache` for `apk add`

### Author/Contributor

### Reviewing Maintainer


sanmai-NL:
@nvuillam ping

nvuillam:
@PeterDaveHello @sanmai-NL Dockerfiles are automatically generated by build.py
So to remove the way apk is called , it must be updated there :)

https://megalinter.io/latest/contributing/#without-write-access

PeterDaveHello:
Thanks @nvuillam, PR updated!

sanmai-NL:
I think there's a better solution, for repeated build runs and for the chance that `apk add` is run more than once.

https://docs.docker.com/build/cache/#use-the-dedicated-run-cache

PeterDaveHello:
Not sure why I should do here 👀

bdovaz:
cc @nvuillam

nvuillam:
let's go :)

PR #2562, opened 2023-04-15
Fix corrective .cspell.json file generated from cspell output

nvuillam:

PR #2673, opened 2023-05-20
feat(linter): add new link checker - lychee

### LINKED ISSUE #2672: feat: new link checker - lychee

I'm always frustrated with `markdown-link-check`. [lychee](https://lychee.cli.rs/) is an alternative that works IMHO!

lychee part of megalinter

N/A

**Additional context**

N/A

DariuszPorowski:
Fixes #2672

## Proposed Changes

1. added new descriptor with [lychee](https://github.com/lycheeverse/lychee)
2. added test fixtures for lychee
3. output files from `build.sh`
4. added https://github.com/oxsecurity/megalinter/raw/main/docs/assets/icons/swift.ico
5. baseline lychee config for ML
6. updated outdated links

### Author/Contributor

### Reviewing Maintainer


bdovaz:
@DariuszPorowski thanks for the contribution.

I have a doubt, and that is why you have created a new descriptor (`lychee.megalinter-descriptor.yml`)? If it is possible, we should fit it in an existing file.

What I don't know is how to categorize it, Maybe `repository`?

cc @nvuillam

nvuillam:
@DariuszPorowski thanks a lot for the PR that looks great :)

I'm aligned with @bdovaz , descriptors are for languages or "file types", and lychee isn't one

For the case where the name linter can be used in several descriptors, it's not very pretty but... we define the linter in both descriptors, like for v8r which is in json and yml

Maybe you could define LYCHEE in md, rst and html descriptors ?

DariuszPorowski:
@bdovaz @nvuillam I struggle with the same case, but cannot figured out the good place 😅 because lychee is not only limited MDs files, and try to avoid code duplication to copy-paste linter descriptor across multiple descriptors. thoughts about the final approach?

echoix:
I have to agree with @DariuszPorowski on this, that copy-pasting isn't ideal. Does the copy-paste linter trigger on this? It would ideally need a comment on the "main" entry to also update the other entries. I didn't know about the v8r duplication, and I don't see a note in the descriptor there to update. Comparing both of the v8r entries, there are the test_folder and examples that are different, plus a different config file name for the yaml entry.

And if the installation instructions are exactly the same, at least it won't be duplicated in the dockerfiles.

However @nvuillam, what would the docs be like if a linter was defined multiple times, in different descriptors? Would we see that the linter is only activated with like, json files for v8r, when we would see another page saying that v8r is activated on yml? I like that when it is duplicated, there are separate configuration variables though, like YAML_V8R_DISABLE_ERRORS_IF_LESS_THAN and JSON_V8R_DISABLE_ERRORS_IF_LESS_THAN

DariuszPorowski:
@bdovaz `repository` (actually that was my initial idea) has `lint_all_files: true` at the root and it does not apply for this case.

@nvuillam if even i duplicate the code for `md`, `rst` and `html` what about the rest of the supported cases for lychee? actually, in my scenarios I use it as we with `yml`, `txt` and `json` :D and `lychee` supports even more formats - do not know it has any limitation (as long as file is kina of text)?

.cspell.json
```diff
@@ -125,6 +125,8 @@
         "Dustilock",
         "Dxml",
         "Dynatrace",
+        "eadme",
```
echoix, line 128: Can cspell use something more complete than part of a regex, since it might be a typo we'd want to catch. 
The original error comes from the regex of the last example in the descriptor:

```yaml
    examples:
      - lychee --format detailed --no-progress README.md info.txt test.html
      - lychee --format detailed --no-progress README.md
      - lychee --format detailed --no-progress test.html info.txt
      - lychee --format detailed --no-progress --offline path/to/directory
      - lychee --format detailed --no-progress https://raw.githubusercontent.com/lycheeverse/lychee/master/README.md
      - 'lychee --format detailed --no-progress "~/projects/big_project/**/README.*"'
      - 'lychee --format detailed --no-progress --glob-ignore-case --verbose "~/projects/**/[r]eadme.*"'
```

nvuillam:
Linters matching N descriptors is a topic for a while in MegaLinter... ( see prettier ! )

I see 3 options:
- CROSSLANG descriptor, where prettier would also go
- LINKS descriptor
- copy-paste in different matching descriptors

The topic is a little bit tricky, because descriptors are not just for "organizing linters", they are also used to create the group of linters performed by a single parallel core (linters that can format/fix files can not run simultaneously, and must be run before the linters of the same descriptors that do not update files)

@bdovaz @Kurt-von-Laven @echoix @DariuszPorowski your opinion ?

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 Note: Can be used with `oxsecurity/megalinter@beta` in your GitHub Action mega-linter.yml file, or with `oxsecurity/megalinter:beta` docker image
 
 - New linters
+  - [#2673](https://github.com/oxsecurity/megalinter/pull/2673) Add Lychee
```
echoix, line 12:
```suggestion
  - [#2673](https://github.com/oxsecurity/megalinter/pull/2673) Add Lychee, a link and email adress checker
```

nvuillam:
See code here about linter groups and sorting by speed -> https://github.com/oxsecurity/megalinter/blob/773fa24b77d5a78cd1732442166b42443d7d48c1/megalinter/MegaLinter.py#L287

megalinter/descriptors/repository.megalinter-descriptor.yml
```diff
@@ -415,3 +422,58 @@ linters:
...
+    install:
+      apk:
+        - wget
+      dockerfile:
+        - |
+          RUN wget --tries=5 -4 -q -O - https://github.com/lycheeverse/lychee/releases/download/v0.13.0/lychee-v0.13.0-x86_64-unknown-linux-musl.tar.gz | tar -xz lychee \
+            && chmod +x lychee \
+            && mv -f lychee /usr/bin/lychee
+        # - FROM lycheeverse/lychee:0.13.0 as lychee
+        # - COPY --link --from=lychee /usr/local/bin/lychee /usr/bin
```
echoix, line 478: Installing from Docker didn't work that's why?

DariuszPorowski: @echoix and this is a great Q... it was my first choice, and always failing.
src of lychee's dockerfile: https://github.com/lycheeverse/lychee/blob/master/Dockerfile

.cspell.json
```diff
@@ -113,7 +113,7 @@
         "DTDs",
         "DUSTILOCK",
         "Dariusz",
-        "Dariusz)    -- megalinter/issues/2448",
+        "Dariusz) -- megalinter/issues/2448",
```
echoix, line 116: Hmm... How come it was needed, it's kind of rare

DariuszPorowski: just restored it from origin, you can see a lot of similarities in cspell, do not know what is the purpose of it. 
image
image

megalinter/descriptors/repository.megalinter-descriptor.yml
nvuillam: This is a ubuntu based image, we probably need an alpine based image to use it as installation instruction

.cspell.json
DariuszPorowski: lets's try remove.

megalinter/descriptors/repository.megalinter-descriptor.yml
echoix: Ah, it's because it's because of the linker used. The Dockerfile used is rust-latest, and the last stage is Debian-based.
Thus, it would be the gnu variant instead of the musl-based target, that would be needed in an Alpine-based container. But they release binaries for all of them, so it's a correct way. You can remove the docker comments for now, since unless they change their docker release workflow, it won't help.
```suggestion
```

DariuszPorowski: @nvuillam they do not release alpine version officially :( https://hub.docker.com/r/lycheeverse/lychee

.cspell.json
echoix: If you have a locally running dev environment with cspell, what does it look like if you exclude instead
```json
        "[r]eadme",
```
or 

```json
        "~/projects/**/[r]eadme.*",
```

DariuszPorowski:
> Linters matching N descriptors is a topic for a while in MegaLinter... ( see prettier ! )
> 
> I see 3 options:
> 
> * CROSSLANG descriptor, where prettier would also go
> * LINKS descriptor
> * copy-paste in different matching descriptors
> 
> The topic is a little bit tricky, because descriptors are not just for "organizing linters", they are also used to create the group of linters performed by a single parallel core (linters that can format/fix files can not run simultaneously, and must be run before the linters of the same descriptors that do not update files)
> 
> @bdovaz @Kurt-von-Laven @echoix @DariuszPorowski Dariusz Porowski FTE your opinion ?

@nvuillam I like CROSSLANG, and I think editorconfig can goes there as well, not only prettier, but it will be breaking change for current user's config

echoix:
> Linters matching N descriptors is a topic for a while in MegaLinter... ( see prettier ! )
> 
> I see 3 options:
> 
> * CROSSLANG descriptor, where prettier would also go
> * LINKS descriptor
> * copy-paste in different matching descriptors
> 
> The topic is a little bit tricky, because descriptors are not just for "organizing linters", they are also used to create the group of linters performed by a single parallel core (linters that can format/fix files can not run simultaneously, and must be run before the linters of the same descriptors that do not update files)

I'm still hesitating because of the parallel groups for concurrency issues. With that point, I understand why duplicating would be advantageous.

.cspell.json
nvuillam: Hmmm this issue was supposed to be solved :'(

nvuillam: https://github.com/oxsecurity/megalinter/pull/2562 😢

nvuillam: If someone here is a regex master, I'd appreciate some help :D

nvuillam:
My preferred option would be "duplicating the linters without copy-pasting YML between descriptors" 🤣 

Something like refs, linter inheritance...

But that would require very impacting updates about how descriptor files are interpreted... I don't feel the courage to go there before V7 :/

.cspell.json
nvuillam: @echoix I think that the most simple if that @DariuszPorowski removes the triggering line in the examples :)

.cspell.json
echoix: @DariuszPorowski Was your branch started locally before that fix, just so we don't work uselessly?

megalinter/descriptors/repository.megalinter-descriptor.yml
echoix: Ya I saw, it's not a big deal neither. They are close to being able to install with binstall, they have everything there. It's not a requirement for megalinter, every environment has different standards.

echoix:
> Something like refs, linter inheritance...

That's what I thought before stopping thinking any further, it would be so complicated to follow along. Something like the descriptor-level `descriptor_flavors:`, but to specify from a linter-level in what descriptor they belong. But it would be a graph based problem to split work afterwards and to show in tables in docs. I'm thinking that if we aren't able to properly show in documentation, it might be too complicated or coupled.

.cspell.json
echoix: These examples are from their docs though.

megalinter/descriptors/repository.megalinter-descriptor.yml
echoix: Oh look, it's already on the way: https://github.com/lycheeverse/lychee/pull/1073
I quickly suggested it in https://github.com/lycheeverse/lychee/issues/59#issuecomment-1555966155

DariuszPorowski:
> I don't feel the courage

`I don't feel the courage` @nvuillam you are French, you have good wines ;) have I solve your courage problem? 😁

.cspell.json
echoix: cspell doesn't take these partial words. So leave as it

Kurt-von-Laven:
> My preferred option would be "duplicating the linters without copy-pasting YML between descriptors" 🤣
> 
> Something like refs, linter inheritance...
> 
> But that would require very impacting updates about how descriptor files are interpreted... I don't feel the courage to go there before V7 :/

This does seem like a lot of risk for one major version update since changing from absolute to relative paths was also high impact. Could be a good goal for v8? What if we did CROSSLANG for now since it seems like significantly less work? We may even be able to leave it at that if the performance impact is not noticeable.

The current level of parallelism trades off some correctness for faster performance since there's no guarantee that the fixes complete before the read-only linters for the same descriptor start to fire. Also, we should be using `len(os.sched_getaffinity(0))` rather than `multiprocessing.cpu_count()` according to [the official Python docs](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.cpu_count). I suspect the consequence of this may be performance degradation from an excessive number of processes on beefy but shared cloud computers (e.g., in CI) where only a small number of cores may actually be used. In fixing these issues, we could also run read-write `CROSSLANG` and `REPOSITORY` linters serially with respect to other linters. I think this may just be Prettier currently, which is fast anyways.

nvuillam:
> > I don't feel the courage
> 
> `I don't feel the courage` @nvuillam you are French, you have good wines ;) have I solve your courage problem? 😁

There is also the time and impact risk issues :D 

@Kurt-von-Laven I still feel that copy-paste in different descriptors would be more relevant and less risky for now (it's just ugly but does not change the docker image build time or size)

Because CROSS_LANG would restrict to linters which can not update files... today we have the "ugly copy paste pattern", maybe let's not create a second one before a big refactoring that would allow to make cleanly all that ^^

echoix:
But once used, and the configuration variables are in the different user's repos, it's a real breaking change to have each user change their config

Kurt-von-Laven:
I get the concern about a breaking config change, although the more I think about this, the more I think we will actually get slightly worse performance with copy-paste than with `CROSS_LANG`. Prettier and misspell are both fast, so it's probably better to launch them once for all file types in series than once per descriptor in parallel. Why would we restrict `CROSS_LANG` to read-only linters though? It's pretty challenging to list all of the file types Prettier and lychee handle. Another pre-existing issue here is that we currently run misspell in parallel with other read-write linters.

nvuillam:
@Kurt-von-Laven we can not have different linters updating the same files at the same time from parallel processes :)

For example, if crosslang_prettier would update markdown files at the same time than markdown_markdown_table_formatter, that would be a mess ☺️

Kurt-von-Laven:
Right, but we could do it serially alongside the bug fix to join the read-write linters before launching the read-only linters. I am imagining a list that we iterate over with a join call at each iteration of the list. The list would contain three elements: the read-write `CROSS_LANG`, `SPELL`, and `REPOSITORY` linters (which I think would just be Prettier and misspell), the other read-write linters grouped by descriptor, and the ungrouped read-only linters.

DariuszPorowski:
oh boy, just wanted to add only one pretty cool linter - it supposed to be easy 😁 so far, do not do any new code updates, before we do not aggrege the approach.

megalinter/descriptors/repository.megalinter-descriptor.yml
DariuszPorowski: @echoix alpine is on the way too https://github.com/lycheeverse/lychee/pull/1074 :D
...it always happened to me 😀, just want to do something what supposed to be quick and easy and finally end up with PRs chain across multiple projects 😅

nvuillam:
@Kurt-von-Laven the queue ordering of group of linters by descriptor (using arbitrary linter_speed attribute: run the longest group of linters first) is quite optimizing the overall execution time, I'm afraid that if we complicate the stuff we could affect performances :/

Each descriptor running of same file type is probably the worst impacting for v7

So... HYPERLINKS descriptor ?
Or maybe even in SPELL one ? (But Spell one is already time consuming so HYPERLINKS descriptor for lychee might be the safer approach

nvuillam:
> oh boy, just wanted to add only one pretty cool linter - it supposed to be easy 😁 so far, do not do any new code updates, before we do not aggrege the approach.

Don't worry, that's with such discussions that we improve MegaLinter from the beginning ^^

And we have much more complex discussions than this one... like the one about multi-arch images for example, that is ongoing for a year 😁

Welcome in MegaLinter community 😅

nvuillam:
@DariuszPorowski someone has to take a decision... and I decide it will be me haha ^^

Go for SPELL descriptor :)

DariuszPorowski:
@nvuillam you say you have ;) I was secretly hoping for HYPERLINKS, because there are still a few solutions on the market and maybe someone will want to integrate :D

Kurt-von-Laven:
Do we want to leave the preexisting race condition where the read-only linters don't wait for the read-write linters to finish alone? If not, I am happy to fix it.

nvuillam:
@DariuszPorowski As I think I'll release v7 tomorrow or sunday, I don't want to take the risk of mess by adding a new descriptor ^^
Anyway, there are pro & cons for all solutions... the best being the reuse of linters between descriptors without copy paste, but this one will be for v8 :D 

@Kurt-von-Laven Lychee does not update files so it will not be meddled with race condition stuff :)

Kurt-von-Laven:
Right, the preexisting bug is completely unrelated to Lychee; I just happened to be reminded of it because I was looking at the concurrency logic for this issue. I don't think that bug fix needs to be tied to a major version release either.

nvuillam:
> Right, the preexisting bug is completely unrelated to Lychee; I just happened to be reminded of it because I was looking at the concurrency logic for this issue. I don't think that bug fix needs to be tied to a major version release either.

Indeed :)

And anyway I don't think it happens often, else someone would have complained ^^

DariuszPorowski:
@nvuillam wait until Sunday, pls :D

nvuillam:
@DariuszPorowski don't worry, I'm pretty sure that not so many people will have tested the beta... so the 7.1.0 should not be much later :D

sanmai-NL:
@DariuszPorowski Would you be interested in applying [`cargo-binstall`](https://github.com/cargo-bins/cargo-binstall) to factor out the downloading and installing logic to a dedicated tool?

nvuillam:
I'd loke to use binstall for all cargo packages, because the existing ones are taking forever to compile during Docker builds :/

DariuszPorowski:
@sanmai-NL @mganter will jump back to this PR later this week to finalize it.
Not sure about Cargo because when I started working on this pr, that option was not available on the lychee side.
What I am sure Docker alpine is now one of the options for lychee.

sanmai-NL:
@DariuszPorowski If it's on crates.io, `cargo install` works. 😊 See https://crates.io/crates/lychee Those other installation options on Lychee's README are listed without awareness of `binstall`.

echoix:
I see that most 404 seems to be network error, like if there wasn't any network access (or permissions?). Do you have similar results by running it locally (and using a custom locally-built image passed to `npx mega-linter-runner --image ...`

I think the ... is the image name , or it might be prefixed with docker.io

DariuszPorowski:
@echoix actually most of them are invalid in real, for example:

https://megalinter.io/configuration/#activation-and-deactivation goes to https://megalinter.io/latest/configuration/#activation-and-deactivation but redirect is on the JavaScript level, not 301 or 302 status code

image

so, it 404 and makes sense.

image

I can exclude the whole domain megalinter.io, but do we want that?

p.s.
network error -> 404 is valid lychee output, not lack of network access, etc

echoix:
Ok, thanks, I see there that the redirect page is a 404.

That was made I think by the limitations that the GitHub hosted pages had when it wasn't a custom domain.

There's probably a way to define the configuration to build the pages better, but @nvuillam is scared to be bitten again by trying to change it

echoix:
What I found weird was the links in the megalinter.io domain having the /raw/ part.

DariuszPorowski:
@echoix `/raw/` is fine because it's 302 redirect to https://raw.githubusercontent.com... but /raw/ with 404 means actually the destination does not exist.
examples:

Bad:
https://github.com/oxsecurity/megalinter/raw/main/docs/assets/icons/swift.ico

Good:
https://github.com/oxsecurity/megalinter/raw/main/docs/assets/images/megalinter-banner.png

DariuszPorowski:
@echoix @nvuillam adjusted lychee config to get rid of false positives as much as possible, and this is my analysis for the rest:

### does not exist (real 404)
- https://github.com/oxsecurity/megalinter/raw/main/docs/assets/icons/security.ico
- https://github.com/oxsecurity/megalinter/raw/main/docs/assets/icons/swift.ico
- https://www.terraform.io/assets/images/logo-hashicorp-3f10732f.svg
- https://black.readthedocs.io/en/stable/editor_integration.html#pycharm-intellij-idea
- https://github.com/sider/goodcheck
- https://stylelint.io/user-guide/integrations/other#analysis-platform-engines
- https://aquasecurity.github.io/trivy/latest/docs/secret/examples/#skip-directories
- https://aquasecurity.github.io/trivy/latest/docs/vulnerability/examples/filter/#by-vulnerability-ids
- https://aquasecurity.github.io/trivy/latest/docs/references/customization/config-file/

### typos - quick fix
- https://pypi.org/project/ansible-lint/'6.16.2' ➡️ https://pypi.org/project/ansible-lint/6.16.2/

### new destination - quick fix
- https://github.com/secretlint/secretlint/raw/main/docs/assets/SecretLintLP.png ➡️ https://github.com/secretlint/secretlint/raw/master/docs/assets/SecretLintLP.png
- https://github.com/stylelint/stylelint/raw/master/identity/stylelint-icon-and-text-white.png ➡️ https://github.com/stylelint/stylelint/raw/main/identity/stylelint-icon-and-text-white.png
- http://mozilla.org/MPL/2.0/ ➡️ https://www.mozilla.org/en-US/MPL/2.0/
- https://www.accurics.com/products/terrascan/ ➡️ https://runterrascan.io/

### JS redirect
- https://pmd.github.io/ ➡️ https://docs.pmd-code.org/ (quick fix)
- https://megalinter.io/ ➡️ https://megalinter.io/latest/ (to discuss)

sanmai-NL:
Eating your own dogfood 🦴 -- in action. 👌🏻

nvuillam:
@DariuszPorowski thanks for the great analysis :)

- https://stylelint.io/user-guide/integrations/other#analysis-platform-engines
  - they seems to have removed this part of the doc, so if we don't find it anywhere 

- https://github.com/oxsecurity/megalinter/raw/main/docs/assets/icons/security.ico & https://github.com/oxsecurity/megalinter/raw/main/docs/assets/icons/swift.ico
  -  we need new icons :D

DariuszPorowski:
@nvuillam @echoix ok, I think it's done :D

nvuillam:
Great PR, great new linter, and great cleaning of broken links, many many thanks @DariuszPorowski 😍

nvuillam (approved):
CLASSY ! :)

DariuszPorowski:
@nvuillam probably it will require minor adjustments overtime for MegaLinter itself, because it's an early-stage baseline config
- lychee config: https://github.com/oxsecurity/megalinter/blob/main/.github/linters/lychee.toml
- lychee ignore: https://github.com/oxsecurity/megalinter/blob/main/.lycheeignore

CC:
@mre / @mre-trv JFYI, your tool is now bundled into MegaLinter :)

nvuillam:
@DariuszPorowski maybe also a default config , we usually provide an simple/opiniotated one in TEMPLATES folder, that is called when no config file is found in the repo :)

nvuillam:
> @nvuillam probably it will require minor adjustments overtime for MegaLinter itself, because it's an early-stage baseline config
> 
> * lychee config: https://github.com/oxsecurity/megalinter/blob/main/.github/linters/lychee.toml
> * lychee ignore: https://github.com/oxsecurity/megalinter/blob/main/.lycheeignore
> 
> CC: @mre / @mre-trv JFYI, your tool is now bundled into MegaLinter :)

@mre I didn't see you're analysis-tools.dev creator, thanks for lychee but also thanks again for analysis-tools.dev, that brings traffic to MegaLinter :)

DariuszPorowski:
@nvuillam what I did wrong? 😅https://github.com/oxsecurity/megalinter/actions/runs/5271203168/jobs/9531767119#step:9:1658

nvuillam:
@DariuszPorowski probably Github repo security because you are not code owner but updated a Github Workflow ^^

And that's good :)

I'll check if it works better once I update the file myself

mre:
> @mre I didn't see you're analysis-tools.dev creator, thanks for lychee but also thanks again for analysis-tools.dev, that brings traffic to MegaLinter :)

@nvuillam, we sort of went full circle with this. 😆 
@DariuszPorowski thanks for your efforts!

High-fives all around.

Issue #2736, opened 2023-06-10
TYPESCRIPT_STANDARD seems to use the vanilla js-standard instead of their recommended ts-standard, causing false positives etc


janderssonse:
**Describe the bug**
When enablling standard linter with typescript, megalinter seems to use the vanilla js(standard) flavour instead of their ts-standard flavour. This will work partially but have bugs for TS-linting, like false positives for no-unused-vars and other rules.

**To Reproduce**
Have a look in the https://standardjs.com/index.html#typescript which recommends the ts-standard, and then in the megalinter docker image which installs standard https://github.com/oxsecurity/megalinter/blob/f27a593354d8787fe2b130fe292f87a55c964088/linters/typescript_standard/Dockerfile#LL90C6-L90C6

For real life test of one these errors with megalinter, create a ts-project, an enum and use the enum variables in other imported classes. Lint with TYPESCRIPT_STANDARD, some things will work but you will get linting errors, as standard can't handle these /ts-standard can).

**Expected behavior**
When enabling TYPESCRIPT_STANDARD ts-standard will be used.

nvuillam:
It probably changed since the creation of MegaLinter :)

Would you like to make the upgrade PR ?

It's probably just cli_executable to force to `ts-standard` in Typescript descriptor :)

janderssonse:
Yes, they created a sublinter-project for the TS-variant due to the differences since then, https://github.com/standard/ts-standard.  . So, started looking at it. it will have to be treated as a new linter basically (as that what it really is TYPESCRIPT_TS_STANDARD should be the natural convention for that take.

janderssonse:
Alrigthee, I had a go, see the pr. Should be a start at least!:)

nvuillam:
@janderssonse can we close the issue now ts-standard is here ? :)

nvuillam:
Plz reopen the issue if you think it deserves to be :)

PR #2746, opened 2023-06-14
feat: add ts-standard linter for ts standard

janderssonse:
This replaces the usage of javascript standard for typescript with the nowadays recommended standard variant ts-standard. It can be seen as a new linter in essence.

Fixes #https://github.com/oxsecurity/megalinter/issues/2736

## Proposed Changes

1. …
2. …
3. …

### Author/Contributor

### Reviewing Maintainer


janderssonse:
I don't expect this to be 100% correct, but maybe it is a start, as it is my first linter try. :) 

Some questions: When I created tests for typescript_ts_standard folder, I still had to duplicate from the test typescript folder. Other tests also seems to do this. Why is that take? How should I do?
Also, why does a lot of tests have same tests duplicated like good_1., good_2, but there is no difference in them.

I also removed references to the Atom editor in the docs for this linter, as that editor is dead.

nvuillam:
@janderssonse great initiative, thanks :)

Duplicating the test folders is when the initial one (with just the descriptor name) that works for most linters do not work with a specific one: in that create we create a test folder just for the linter

About multiple good or bad tests, it's to test the capability of the linters to accept a list of files and not a single file :)

janderssonse:
@nvuillam I didn't see the use case for a specific typescript_ts_standard test folder after the test explanation, so that is gone now. PR updated. No hurries, when you get the time to look at it, please also tell me what I did wrong with this one so next linter adventure can improve. I might have some further PR's for the project in mind, but let's get this one right first:)

nvuillam:
@janderssonse looks good, but please can you merge main in your branch &  solve conflicts ? :)

nvuillam:
(you can ignore the issue about the new doc page not found)

nvuillam (changes requested):
The content seems very good, except one think: you can not change the key of the linter (you can force it with `name` property in the descriptor if necessary), else it would require a deprecation then a major release.

If ts-standard is done by the same team than standard, and works the same, the linked key can remain unchanged :)

janderssonse:
> The content seems very good, except one think: you can not change the key of the linter (you can force it with `name` property in the descriptor if necessary), else it would require a deprecation then a major release.
> 
> If ts-standard is done by the same team than standard, and works the same, the linked key can remain unchanged :)

Another go, now it is called TYPESCRIPT_STANDARD (not ...TYPESCRIPT_TS_STANDARD after some quick replaces, as it was originally called

nvuillam:
Thanks for the update, and for the whole PR :)

PR #2805, opened 2023-07-13
⬇️ fix(requirements.txt): downgrade pymdown-extensions from 10.0 to 9.11

ThomasSanson:
The pymdown-extensions package was downgraded to version 9.11 due to compatibility issues with other packages in the project. The newer version was causing conflicts, hence the need for the downgrade.

### DIFF
--- a/.config/python/dev/requirements.txt
+++ b/.config/python/dev/requirements.txt
@@ -16,7 +16,7 @@ mkdocs-material
 multiprocessing_logging
 pychalk
 pygithub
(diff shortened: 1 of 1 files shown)


ThomasSanson:
> I just added grype so CI is broken, but we'll fix that ;)

giphy51

🤣

nvuillam (approved):
Thanks for your reactivity @ThomasSanson :)
Let's hope it works ^^

Issue #2834, opened 2023-07-25
trufflehog linter raises spurious false positives


quotidian-ennui:
**Describe the bug**

New 'trufflehog' linter causes issues out of the box since it detects potential secrets inside the ".git" directory if you run megalinter as a github action.

**To Reproduce**

1. Install and use megalinter as a github action; with `pull_request: write` and `contents:write` permissions
2. Trigger it.

You will intermittently get things like

```
❌ [ERROR] for workspace /github/workspace
Linter raw log:
2023-07-25T16:20:29Z	info-0	trufflehog	loaded decoders	{"count": 3}
2023-07-25T16:20:29Z	info-0	trufflehog	loaded detectors	{"total": 748, "verification_enabled": 748, "verification_disabled": 0}
🐷🔑🐷  TruffleHog. Unearth your secrets. 🐷🔑🐷

Found unverified result 🐷🔑❓
Detector Type: Github
Decoder Type: BASE64
Raw result: ghs_xxxxxxx
File: .git/config
Line: 11
```

- This is potentially correct but still a false positive because there's not an awful lot you can do about it... (since I imageine that this is under the control of actions/checkout)

```
Results of trufflehog linter (version 3.45.0)
See documentation on https://megalinter.io/7.2.0/descriptors/repository_trufflehog/
-----------------------------------------------

❌ [ERROR] for workspace /tmp/lint
Linter raw log:
2023-07-25T15:03:12Z	info-0	trufflehog	loaded decoders	{"count": 3}
2023-07-25T15:03:12Z	info-0	trufflehog	loaded detectors	{"total": 748, "verification_enabled": 748, "verification_disabled": 0}
🐷🔑🐷  TruffleHog. Unearth your secrets. 🐷🔑🐷

Found unverified result 🐷🔑❓
Detector Type: Github
Decoder Type: PLAIN
Raw result: -some-hash-here
File: .git/logs/HEAD
Line: 29

Found unverified result 🐷🔑❓
Detector Type: Github
Decoder Type: PLAIN
Raw result: -some-hash-here
File: .git/logs/HEAD
Line: 62
```

- Which is quite obviously a false positive.

nvuillam:
Indeed that's strange...

For the moment I didn't find a way to ignore false positives in trufflehog, but I'll check again !

Meanwhile you can just disable errors by adding

```yaml
DISABLE_LINTERS_ERRORS:
  - REPOSITORY_TRUFFLEHOG
```

[...]

PR #2838, opened 2023-07-26
Call trufflehog with `--only-verified` to avoid false positives in .git/config

nvuillam:
Fixes https://github.com/oxsecurity/megalinter/issues/2834


quotidian-ennui (commented):
Should you re-enable trufflehog in your .mega-linter.yml so that you start "drinking your own champagne" again 😉

PR #2891, opened 2023-08-17
Reformatted config files to not fail common linters (closes #2890)

### LINKED ISSUE #2890: Generated workflow file fails most default linting tools
**Describe the bug**
As-is, the default `mega-linter.yml` file fails common linting tools like Editorconfig checker, especially if it checks line length. This can be resolved with better formatting of instructive comments and the use of blocks for `if` conditionals instead of having them all on one line.

Additionally, the job name should be something less generic than `build` - these names are often used for referencing specific jobs for individual testing. Naming it something like `megalinter` would cause less conflict in these instances. **I do not know whether this would be considered a breaking change, given the build step of Megalinter changes, and this might disrupt testing scripts (e.g., I use `act`).**

**To Reproduce**
`npx mega-linter-runner --install`

**Expected behavior**
Common linting tools, such as Editorconfig Checker, should pass.

andrewvaughan:
Fixes #2890

_**Important**_ - I am still unable to run tests locally, so a solid code-review would be very-much appreciated.

Additionally, I believe that the change of the job name from `build` to `megalinter` - while extremely important for people who build automation scripts that reference the build name (like myself) and need something more specific than "build", this may end up being a breaking change if anyone has built such scripts or dependencies in the past. Please keep this in mind for release policies.

## Proposed Changes

1. Used more-advanced YAML to better-format config files so they don't fail with common linters (especially those that enforce line-length)
2. Changed the workflow job name from a generic `build` which can conflict with other workflows to `megalinter`

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 Note: Can be used with `oxsecurity/megalinter@beta` in your GitHub Action mega-linter.yml file, or with `oxsecurity/megalinter:beta` docker image
 
+- Workflow job name changed from `build` to `megalinter` to prevent conflicts with other workflows
 - Linter versions upgrades
 - Secretlint logo - reduce size to 150 and remove background
   - [mypy](https://mypy.readthedocs.io/en/stable/) from 1.4.1 to **1.5.0** on 2023-08-11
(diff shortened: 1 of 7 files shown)


nvuillam (approved):
Looks good to me, thanks a lot for your contribution :)

nvuillam:
@andrewvaughan it would reassure me if you can test a little with `npx mega-linter-runner@beta --install` before we release a new version :) ( otherwise I'll test it myself but I have some time availability issues these days ^^ )

PR #3067, opened 2023-11-03
C flavor

daltonv:
## Proposed Changes

Adds a new c_cpp flavor that can later be used by clang-format (see #763)

### Author/Contributor

### Reviewing Maintainer


daltonv:
I will still add clang-format. I just wanted to keep my first PR simple as this is my first contribution.

megalinter/descriptors/c.megalinter-descriptor.yml
```diff
@@ -2,6 +2,7 @@ descriptor_id: C
 descriptor_type: language
 descriptor_flavors:
   - cupcake
+  - c_cpp
   - dotnet
```
daltonv, line 6: I still see a problem here when it comes to `clang-format` being added. I would naturally put that linter in here and the cpp descriptor file.

However, if I understand it right, that would mean `clang-format` is also added to the `cupcake`, `dotnet`, and `dotnetweb` linters. I think we only want it in this new `c_cpp` linter. Right?

nvuillam: @daltonv you can override descriptor_flavors at linter level, so you could add  the following property on clang-format linter definition :)

```
     descriptor_flavors: 
        - c_cpp
```

nvuillam (approved):
Looks very nice to me, many thanks for your contribution :)

PR #3075, opened 2023-11-04
Fix csharp installation dependencies

nvuillam:


bdovaz:
@nvuillam why this PR? What problemas did you solve? There is no reference to any issue

PR #3126, opened 2023-11-18
Fix powershell installation

nvuillam:
Use official powershell installation instructions

https://learn.microsoft.com/en-us/powershell/scripting/install/install-alpine?view=powershell-7.4


bdovaz:
@nvuillam why did you change the installation mode? The previous one (using API) allowed you to automatically reference "latest", now it is not possible and you have to change it manually, i.e. now it is more expensive to keep powershell updated because it is spread over several different descriptors.

cc: @echoix @Kurt-von-Laven

nvuillam:
Usually when i do an update like that it's because Auto-update-linters job is failing with its current instructions , that probably was the case :)

PR #3155, opened 2023-11-23
Add Roslynator linter

bdovaz:
Fixes https://github.com/oxsecurity/megalinter/issues/2269

## Proposed Changes

**Dotnet restore command:**

In the end I put it in the "process_linter" method because it is the only one I see that I could do it in.

**Related to tests:**

I had to add 2 specific variables to the descriptors for the format_fix test.
- **test_format_fix_file_extensions**: specific extensions to this test
- **test_format_fix_regex_exclude**: exclusions specific to this test.

You may ask, why?

Well, because this linter takes as input ".csproj" files but when doing format/fix it actually modifies the ".cs" files which are the scripts contained in the project (.csproj).

The value that I have put to each variable is the following:
- test_format_fix_file_extensions: .cs -> so that the diff that does the format_fix test only takes into account the files it actually alters.
- test_format_fix_regex_exclude: /bin/|/obj/ -> as it has to run a `dotnet restore` previously, it generates the folders /bin, /obj and these folders do not have to be taken into account.

Besides in this test I have simplified the file filtering logic and I have used the "core" `utils.filter_files` method to avoid duplicated and error prone logic


bdovaz:
@nvuillam do you know what problem I may have here:

https://github.com/oxsecurity/megalinter/blob/f680b936b4d171632ece4b08090fbf6cacc80a0b/megalinter/utilstest.py#L634

What causes it not to filter files that it should?

https://github.com/oxsecurity/megalinter/actions/runs/6997890867/job/19035355401?pr=3155#step:9:294

Thank you!

nvuillam:
@bdovaz maybe your updates in utils_test.py ?

image

bdovaz:
Either you didn't understand me or I didn't explain well.

I know that the problem is in that file but what I mean is that I have changed that instead of using a custom code I use the method:

utils.filter_files

But there is some input that I am passing it wrong because it returns files that it should not as I have put in the second link.

As I imagine that you created that method that's why I ask you if you saw at a glance something wrong.

Thanks!

echoix:
I don't remember exactly, but I followed the imports and read the code, and searched on google in parallel. I saw the offending code. I found the specific issue on github the other month, but lost track of it.

bdovaz:
@nvuillam ready for you to review!

The builds that fail are because of the same thing:

image

Complaining about a file that evidently does not yet exist in the main branch.

nvuillam:
@bdovaz thanks for this great addition 😍😍

PR #3225, opened 2023-12-16
Upgrade actions/upload-artifact@v3 to actions/upload-artifact@v4 in default workflows

nvuillam:


echoix (approved):
As long as there are no duplicated artifacts name in a same job, it should work!

PR #3318, opened 2024-01-24
Format powershell linter output into terminal-wide table

### LINKED ISSUE #3313: Improve output format of the powershell linter
The output of the powershell linter is formatted as an 80 chars-wide table. This makes the describing message difficult to read, but will also break the content of the `RuleName` and `ScriptName` columns into several lines.

Output should adapt to the size of the terminal.

I think that the only required changed is to pipe the call to `Invoke-ScriptAnalyzer` into the `Format-Table` cmdlet, as in the example below:

```powershell
Invoke-ScriptAnalyzer -EnableExit -Path path-to-file.ps1 | Format-Table -AutoSize -Wrap
```

**Additional context**

efrecon:
Fixes #3313 and #3320

## Proposed Changes

Improve the output format of the `powershell` linter by ensuring that the columns utilise the entire terminal width. This improves readability of both the description message and the location of the offending problem. Extra care is necessary to bypass a known powershell [bug](https://github.com/PowerShell/PowerShell/issues/20110) on UNIX-like systems.

### Author/Contributor

### Reviewing Maintainer


efrecon:
I see that the tests [fail](https://github.com/oxsecurity/megalinter/actions/runs/7641033932/job/20826142196?pr=3318#step:11:3198). This is because the output has changed and is not clamped to 80 characters anymore.

An "easy" solution would be to arrange for the content of the [ERROR-POWERSHELL_POWERSHELL.txt](https://github.com/oxsecurity/megalinter/blob/main/.automation/test/powershell/reports/ERROR-POWERSHELL_POWERSHELL.txt) file to be the same as the one when running the tests at GitHub. Where can I find the output in the GitHub workflow logs/files?

A better solution, i.e. working in all cases, would require to add a new environment variable called `POWERSHELL_POWERSHELL_OUTPUT_WIDTH` (or similar) and to:
1. Control and enforce the content of that variable during the tests.
2. Arrange for the output file to match.

Which solution is best and how can we enforce the content of an environment variable during a test?

nvuillam:
The error seems related to the fact that the terminal is not tty (basically means that it is not interactive)

```sh
        if self.linter_name == "powershell":
            # Format output to fit in terminal, respecting the terminal width.
            # For more info: https://stackoverflow.com/a/76889369
>           size = os.get_terminal_size()
E           OSError: [Errno 25] Not a tty
```

You should implement a fallback with a default size, or even just define a default size, as terminal will never be interactive,  and it should work :)

nvuillam:
sorry for the delay, i'm very busy these day ^^
Thanks for the new commits, let's hope CI jobs will pass :)

nvuillam (approved):
thanks a lot @efrecon  for your hardwork on this PR :)

PR #3430, opened 2024-03-15
Ensure trufflehog does not auto-update itself

wandering-tales:
It looks like that the [trufflehog](https://github.com/trufflesecurity/trufflehog) linter auto-updates itself to the latest version when run. That makes the pinning of its version ineffective and causes builds to be non deterministic.

I have discovered such behavior by suddenly stumbling upon a false positive, probably due to a [change](https://github.com/trufflesecurity/trufflehog/pull/2372) introduced in the latest release [v3.70.0](https://github.com/trufflesecurity/trufflehog/releases/tag/v3.70.0).

## Proposed Changes

1. Ensure trufflehog does not auto-update itself

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Note: Can be used with `oxsecurity/megalinter@beta` in your GitHub Action mega-l
 
 - Fixes
   - Fix conflict between prettier and yamllint about spaces
+  - Ensure [trufflehog](https://github.com/trufflesecurity/trufflehog) does not auto-update itself
 
 - Doc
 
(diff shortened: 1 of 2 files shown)


nvuillam (approved):
Some security linters update their database, but I'm against updating the whole executable version when not requested... so good catch, thanks for your contribution :)

wandering-tales:
You're welcome :)

Anyway, just for reference, here's a gist including a GitHub Action log excerpt showing that latest version of trufflehog (v3.70.0) was used instead of the one pinned in Megalinter 7.10.0 (v3.69.0):

https://gist.github.com/wandering-tales/51311d13eccf03a3b27388069dbcd0a4

PR #3476, opened 2024-04-08
fix: remove `TEMPLATES/.secretlintignore`

### LINKED ISSUE #3328: secretlint documentation bug re: no .secretlintignore file
**Describe the bug**
I was wondering about the behavior of `secretlint` and looked at #2992. It led me to see that there is a difference between implementation and documentation. 

The documentation at https://megalinter.io/latest/descriptors/repository_secretlint/ says:
> See [How to ignore files and directories with secretlint](https://github.com/secretlint/secretlint/blob/master/docs/configuration.md#secretlintignore)
> If custom `.secretlintignore` ignore file is not found, [`.secretlintignore`](https://github.com/oxsecurity/megalinter/tree/main/TEMPLATES/.secretlintignore) will be used

**Expected behavior**
To me, this implies that if I do not configure a `.secretlintignore` file, then a default one will be generated on the fly with the same contents as at that link, then the file will be discarded.

**Actual behavior**
However, [the actual code](https://github.com/oxsecurity/megalinter/blob/b2813b52bde3fb7c236ede8e64b257107a0c4a58/megalinter/linters/SecretLintLinter.py#L30) shows that if a custom `.secretlintignore` file is not found, then `.gitignore` will be used, not a default.

**Desired outcome**
Update the documentation to match the implementation.

pjungermann:
Removes the default `.secretlintignore`
that was actually never applied.

If `.secretlintignore` is absent,
`.gitignore` is used instead.

Removing this default config will adjust
the misleading documentation as well.

Fixes #3328

## Proposed Changes

1. …
2. …
3. …

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Note: Can be used with `oxsecurity/megalinter@beta` in your GitHub Action mega-l
   - Salesforce linters: use sf + default Flow Scanner rules
   - Disable JSON_ESLINT_PLUGIN_JSONC until <https://github.com/ota-meshi/eslint-plugin-jsonc/issues/328> is fixed
   - Upgrade tar in mega-linter-runner
+  - secretlint: remove default `.secretlintignore` that was never used but `.gitignore` is used instead. Fixes [#3328](https://github.com/oxsecurity/megalinter/issues/3328)
 
 - Doc
   - Update R2DevOps logo
--- a/TEMPLATES/.secretlintignore
+++ b/TEMPLATES/.secretlintignore
@@ -1,1 +1,0 @@
-megalinter-reports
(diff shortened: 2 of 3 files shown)


echoix (commented):
I'm thinking about the use case where people use mega-linter-runner to run Megalinter on a new repo or ad-hoc. Is there a way where megalinter-reports could be excluded? In this use case, megalinter-reports probably isn't in their .gitignore file yet.

Maybe in the changelog, add the note to make sure megalinter-reports folder is in their .gitignore file or something like that, with the correct info?

pjungermann:
@echoix Sure, I can add this recommendation to the changelog.

However, the removed file was never used. Even before this change, it used the `.gitignore` file. Only that the documentation suggested something else.

See also the linked issue #3328

echoix (approved):

PR #3647, opened 2024-06-10
chore(deps): upgrade dependencies yeoman-environment to 4.4.1 and yeoman-generator to 7.2.0

### LINKED ISSUE #3623: Dependency "ip" needs to be replaced
As described in #3619 we need to update the transitive dependency `ip` to a version without a known CVE or get rid of it completely.

Afterwards, e7a61250d89d14ebc61ea9c3547d442c2b5136d4 and 8ae5f2d78805bc470e7c0968a029f74771d7aef4 need to be reverted.

muffl0n:
Fixes #3623


echoix (changes requested):
Need to find where ip is really installed

PR #3664, opened 2024-06-16
enhancement: CSharpier config filename and path override ability

### LINKED ISSUE #3553: Specify Configuration File for CSharpier Linter
Currently, there are no environment variables available for CSharpier linter to point to a configuration file, making it challenging to specify rules from a specific configuration file. This limitation inhibits users from customizing linting rules efficiently.Hence, with current implementation, how can we set CSharpier linter to point to a custom configuration file?

TommyE123:
Fixes #3553

# [CSharpier](https://csharpier.com) - Filename and path override improvements
This change adds the ability to override the location of the `.csharpierrc ` filename and path. Currently its using the default option for the command `-c` which is incorrect and requires `--config-path ` [csharpier documentation](https://csharpier.com/docs/CLI).
This also now allows the `.csharpierrc ` file to be placed in the default linter path which currently doesn't work due to the wrong command parameter.

Please note this doesn't allow an `.editorconfig` to work in another location as this isnt currently supported by csharpier. (see [issue 3553](https://github.com/oxsecurity/megalinter/issues/3553) for more info) However if this changes then the fix should also work nicely with this approach as well.

## Proposed Changes

1. Fixed config path option command.
2. Added default file name .csharpierrc.
3. Updated schema to reflext the 2 new available options avaiable. (assumed this was required but please say if this isn't or I need to update anything else?)

## Testing - Mixture of CI builds to demonstrate the fix

**GITLAB - Both filename and location overridden:**
image
image

**GITLAB - .csharpierrc in default Linter rules location:**
image

**AZURE - Filename overridden .csharpierrc.yml in default Linter rules location:**
image
image

**GITHUB - Filename .csharpierrc.yml in root (no extra megalinter parameters set):**
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Note: Can be used with `oxsecurity/megalinter@beta` in your GitHub Action mega-l
 
 - Linters
   - Add python package Pygments to rst-lint venv
+  - [CSharpier](https://csharpier.com) added ability to override config filename and path
 
 - Reporters
 
--- a/megalinter/descriptors/csharp.megalinter-descriptor.yml
+++ b/megalinter/descriptors/csharp.megalinter-descriptor.yml
@@ -45,6 +45,8 @@ linters:
     linter_rules_configuration_url: https://csharpier.com/docs/Configuration
     linter_rules_ignore_config_url: https://csharpier.com/docs/Ignore
     ignore_file_name: .csharpierignore
+    config_file_name: .csharpierrc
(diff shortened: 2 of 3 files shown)


echoix (approved):
Again as always, you went above and beyond to prove the work and the justifications. (I don't even have access to a setup that has these three CI platforms). The actual contents of the PR look correct.

Rerunning the failed test for unrelated reasons, and setting up the automerge like the other PR.

PR #3978, opened 2024-09-09
feat(descriptors, lua): add `selene` support

AlejandroSuero:
Added `selene` lua linter support to `lua` descriptors.
#3971

Fixes #

## Proposed Changes

1. Added [selene](https://github.com/Kampfkarren/selene) to descriptors

### Author/Contributor

### Reviewing Maintainer


echoix:
@nvuillam and what is needed as test file for a new linter?

echoix:
Since updating the repo from the new linter modifies some workflow files, I need a human to run the build command.

If you are on Linux, it is `./build.sh`, otherwise, 

```shell
pip install pipx
pipx install hatch
hatch shell

hatch run build:build
```

nvuillam:
> @nvuillam and what is needed as test file for a new linter?

About tests, let's see if the ones existing for luacheck do the work :)

https://github.com/oxsecurity/megalinter/tree/main/.automation/test/lua

nvuillam:
@AlejandroSuero , I think you can factorize the installation of lua at descriptor level (`install` property also defined at top level )

And keep at linter level only what is purely related to the linter :)

Example with powershell: https://github.com/oxsecurity/megalinter/blob/main/megalinter/descriptors/powershell.megalinter-descriptor.yml

AlejandroSuero:
@nvuillam here are the results for selene in existing tests:

Screenshot 2024-09-09 at 19 38 55

nvuillam:
@AlejandroSuero can you update the test cases so they work with both luacheck and selene ?

Otherwise, it's possible to create a test cases folder just for lua_selene and set property test_folder in selene descriptor, but if we can avoid that I prefer :)

AlejandroSuero:
> @AlejandroSuero , I think you can factorize the installation of lua at descriptor level (`install` property also defined at top level )
> 
> And keep at linter level only what is purely related to the linter :)
> 
> Example with powershell: https://github.com/oxsecurity/megalinter/blob/main/megalinter/descriptors/powershell.megalinter-descriptor.yml

On the `powershell` one the top-level `install` it installs powershell itself, but then the installs on the linters are separated from that.

Do I move the `luacheck` install and `selene` install into the same one in the top-level?

echoix:
> @AlejandroSuero , I think you can factorize the installation of lua at descriptor level (`install` property also defined at top level )
> 
> 
> 
> And keep at linter level only what is purely related to the linter :)
> 
> 
> 
> Example with powershell: https://github.com/oxsecurity/megalinter/blob/main/megalinter/descriptors/powershell.megalinter-descriptor.yml
> 
> 

The next linter is a Docker copy-only linter though, so in individual linter the language-level install will be suboptimal

echoix:
Do you want to rerun the same build commands (when changing the descriptors, so the Dockerfiles are updated)? I can try to run them here instead, but you'll have to pull after (and I'll close and reopen the PR to trigger CI after the push).

AlejandroSuero:
@echoix I ran the command. Do I commit&push them now or do I wait for the reopen to pull them?

echoix:
I would need to close and reopen only if it is the GitHub bot in CI (that used the GITHUB_TOKEN) that pushed the changes. Since it is a human, you can push it as you want (so the correct Dockerfiles are built and tested in CI).

You're doing great!

megalinter/descriptors/lua.megalinter-descriptor.yml
```diff
@@ -2,6 +2,14 @@ descriptor_id: LUA
...
 file_extensions:
   - ".lua"
+install:
+  dockerfile:
+    - |
+      RUN wget --tries=5 https://www.lua.org/ftp/lua-5.3.5.tar.gz -O - -q | tar -xzf - \
+          && cd lua-5.3.5 \
+          && make linux \
+          && make install \
+          && cd .. && rm -r lua-5.3.5/ \
```
echoix, line 12:
```suggestion
          && cd .. && rm -r lua-5.3.5/
```
After that, rebuild (sync) the Dockerfiles and should fix the invalid build

echoix (changes requested):

megalinter/descriptors/lua.megalinter-descriptor.yml
```diff
@@ -54,3 +57,37 @@ linters:
...
+    cli_config_arg_name: "--config"
+    examples:
+      - "selene myfile.lua"
+      - "selene --config chktexrc.toml myfile.lua"
+    install:
+      apk:
+        - g++
+      dockerfile:
+        - |
+          RUN cargo install --branch main --git https://github.com/Kampfkarren/selene selene
```
echoix, line 79: We try to use official releases, so maybe try the following:

```suggestion
          RUN cargo install selene@0.27.1
```

echoix (changes requested):

megalinter/descriptors/lua.megalinter-descriptor.yml
echoix, line 79: @nvuillam where does cargo get set up in individual linters? As it isn't being built in the PR, we don't see it fail yet

echoix: Im still trying to find the correct syntax to keep the version number updated

echoix, line 79: I just remembered we have some handling for cargo crates directly

```suggestion
      apk:
        - g++
      cargo:
        - selene
```

Or if g++ isn't needed anymore

```suggestion
      cargo:
        - selene
```


If we support the version number correctly:

```suggestion
      cargo:
        - selene@0.27.1
```

echoix (changes requested):

megalinter/descriptors/lua.megalinter-descriptor.yml
nvuillam: Indeed, if there is a single cargo package in the dockerfile, rust toolchain commands are automatically added before :)

echoix: @nvuillam  Do we support having docker env vars as the cargo version when building our Dockerfiles? Like

```yaml
cargo:
  - selene@${LUA_SELENE_VERSION}
dockerfile:
  - |-
    # renovate: datasource=crate depName=selene
    ARG LUA_SELENE_VERSION=0.27.1
```

Would the arg be placed before using the cargo to install?

AlejandroSuero: For what I saw with `cargo` in the rust descriptor, it's just being omitted (just using the crate name):

https://github.com/oxsecurity/megalinter/blob/8ec5e3ef2d860b1bdf705e60f0e38a36d75c3e55/megalinter/descriptors/rust.megalinter-descriptor.yml#L23-L25

echoix: We weren't pinning packages up until a couple weeks ago, from my push to make keeping them updatable manageable. So you'll find lots of places where they aren't implemented yet

AlejandroSuero: Do I leave it with `selene@0.27.1` or just `selene` for the time being.

echoix: We can try with this: https://github.com/oxsecurity/megalinter/pull/3978#discussion_r1750895746

AlejandroSuero: @echoix LGTM 👍

echoix:
I'm running tests of individual linters in https://github.com/echoix/megalinter/pull/120

megalinter/descriptors/lua.megalinter-descriptor.yml
```diff
@@ -22,12 +30,7 @@ linters:
         - readline-dev
```
echoix, line 30: Ok, so after the tests of building individual linter images in my fork, I found out that this readline-dev is a dependency of lua itself, but openssl is only required for a library used by luacheck, which uses either that, md5sum or md5. 

So, this needs to be moved to the language-level install, but openssl kept here.

echoix (changes requested):

echoix (approved):

echoix:
Thanks again! I assure you, the StyLua is way easier to integrate, it's only a copy operation

nvuillam:
Great job @AlejandroSuero , i'll mention you in the LinkedIn post at the next MegaLinter release 😊

AlejandroSuero:
@nvuillam @echoix thanks for the feedback and the help.

image

Should this be the way to copy it?

https://github.com/JohnnyMorganz/StyLua?tab=readme-ov-file#docker

https://hub.docker.com/r/johnnymorganz/stylua/tags

echoix:
Yes, but I think there's missing the $ for using the var

PR #4170, opened 2024-10-20
Squash docker images to have less layers and size

nvuillam:


echoix:
Is it possible to not squash all layers at once? Pulling performance will be worse, as it would need to wait to complete the complete image before starting to verify, then extract. Any change will need to have all contents changed.

It's possible to change some dockerfiles to have multiple layer operations combined together by using multistage builds and copying the result of multiple layers as one layer. I don't know if there are other ways to not end up with that big squash.

echoix:
And size-wise, it's true when you waste space by deleting in another layer, or changing the same file later. Last time I checked, we were pretty good at that, and we weren't wasting much. Pretty much some tiny files that were touched by tools that we couldn't remove.

nvuillam:
@echoix i don't have other ideas when I see simple docker pulls fail, like this one :/

https://github.com/oxsecurity/megalinter/actions/runs/11427413336/job/31791515470

echoix:
> @echoix i don't have other ideas when I see simple docker pulls fail, like this one :/
> 
> https://github.com/oxsecurity/megalinter/actions/runs/11427413336/job/31791515470

I tried it locally (on the gitpod instance I was using, not for megalinter), and get this too. So there's something else going on.

image

echoix:
https://stackoverflow.com/questions/47272611/docker-max-depth-exceeded

nvuillam:
On a new action I don't see anything I could call to avoid such error, that's why i try the squash :/

echoix:
Ah see that the overlay2 storage engine has a limit of 125 layers: https://github.com/docker/for-linux/issues/414#issuecomment-438861366

But how we end up to 125 is probably there the problem. A recursive something propably happened in the chain somewhere, that is adding layers upon layers

echoix:
It seems that containerd backing store can handle these. So it is a possibility to use that for debugging what is causing so many layers for a worker image

PR #4530, opened 2025-01-19
Use renovate with clj-kondo

bdovaz:
For determinism when building Docker images, it is important that we do not use the master branch directly.


bdovaz:
Ready!

cc @nvuillam @echoix

nvuillam (approved):

PR #4531, opened 2025-01-19
Use renovate with clj-style

bdovaz:
For determinism when building Docker images, it is important that we do not use the main branch directly.


bdovaz:
Ready @nvuillam!

nvuillam (approved):

PR #4533, opened 2025-01-19
Use renovate with dart

bdovaz:
In this case, it had a really old fixed version


megalinter/descriptors/dart.megalinter-descriptor.yml
```diff
@@ -68,7 +68,8 @@ linters:
       ## stable dart sdk: https://dart.dev/get-dart#release-channels
       ## install alpine-pkg-glibc (glibc compatibility layer package for Alpine Linux)
       dockerfile:
-        - ARG DART_VERSION='2.8.4'
+        # renovate: datasource=dart
```
echoix, line 71: There's two required fields in that regex.

echoix (changes requested):

megalinter/descriptors/dart.megalinter-descriptor.yml
bdovaz: So what do I have to put `depName=dart`?

echoix: I took this https://docs.renovatebot.com/presets-customManagers/#custommanagersdockerfileversions, and adapted it to allow changing our descriptor files too. Plus adapted our build scripts to keep the comment together with the arg when writing dockerfiles

More understandable docs: 
https://docs.renovatebot.com/modules/manager/regex/

echoix: Renovate handles a lot of things, formats, sources, etc. But at some point, it needs to know where to look, and what to look for. In a custom manager, it can't know where to look, unlike if it was a npm package.json, it would look in the npm registry.

bdovaz: @echoix as far as I understand the correct datasource is: https://docs.renovatebot.com/modules/datasource/dart-version/

And looking at this example we can use `depName=dart` as I said: https://github.com/renovatebot/renovate/discussions/26761#discussioncomment-8641152

megalinter/descriptors/dart.megalinter-descriptor.yml
```diff
@@ -68,7 +68,8 @@ linters:
       ## stable dart sdk: https://dart.dev/get-dart#release-channels
       ## install alpine-pkg-glibc (glibc compatibility layer package for Alpine Linux)
       dockerfile:
-        - ARG DART_VERSION='2.8.4'
+        # renovate: datasource=dart-version depName=dart
+        - ARG DART_VERSION='3.6.1'
```
echoix, line 72: Seems only like a comment in the descriptor, not copied over to the Dockerfiles

echoix (changes requested):

megalinter/descriptors/dart.megalinter-descriptor.yml
bdovaz: Done!

echoix (approved):

echoix:
Version of dartanalyzer doesn't seem properly captured.

Also: 

Fatal error while calling dartanalyzer: [Errno 2] No such file or directory: 'dartanalyzer'


Did the binary name change since the last very old version?

bdovaz:
> Version of dartanalyzer doesn't seem properly captured.
> 
> Also:
> 
> Fatal error while calling dartanalyzer: [Errno 2] No such file or directory: 'dartanalyzer'
> 
> Did the binary name change since the last very old version?

It does not seem so: https://storage.googleapis.com/dart-archive/channels/stable/release/3.6.1/sdk/dartsdk-linux-x64-release.zip

bdovaz:
> Version of dartanalyzer doesn't seem properly captured.
> 
> Also:
> 
> Fatal error while calling dartanalyzer: [Errno 2] No such file or directory: 'dartanalyzer'
> 
> Did the binary name change since the last very old version?

It seems that, in fact, it no longer exists:

image

bdovaz:
Ready @nvuillam!

In the end it took me more work because of the jump from 2.x to 3.x.

bdovaz:
I have taken the opportunity to change urls and make it work to get the version of the linter

PR #4534, opened 2025-01-19
Use renovate with ktlint

bdovaz:
For determinism when building Docker images, it is important that we do not use the main branch directly.


bdovaz:
Ready!

cc @nvuillam @echoix

echoix (approved):

nvuillam (approved):

PR #4707, opened 2025-02-11
Better apk package url

bdovaz:
Generate the url by setting the following parameters:
* **branch**: points to the correct alpine version and not to `edge`
* **arch**: points to the correct architecture instead of showing all of them


bdovaz:
Ready to review!

cc @nvuillam @echoix

nvuillam (approved):

PR #4932, opened 2025-03-13
Update plugins.md to add raw link to JSON schema

wesley-dean:
This adds a link to the "raw" version of the JSON schema for descriptors.

Fixes #

## Proposed Changes

1. this adds a link labeled`(raw)` after the reference to the JSON schema for linter descriptors.  While working on validating a plugin, I attempted to use `v8r` with the URL mentioned on the plugins.md page only to be met with the error `Unexpected token '<', "` which confused me as there was no `<` in my plugin descriptor.  I eventually realized that the link on the plugins page points to the JSON schema file in the repo, not raw version that I could pass to `v8r -s https://....` .  Rather than change the existing link, I added a new link with the `(raw)` title after it.  I'm pretty sure there are more attractive ways to accomplish the same thing, so I'm very open to feedback, modifications, etc..

### Author/Contributor
- [ ] Add entry to the CHANGELOG: I didn't because it's just a small addition to a page on the site, not MegaLinter itself
- [ ] If documentation is needed: same, just a small update to the site

### Reviewing Maintainer

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -47,7 +47,7 @@ PLUGINS:
 You can implement your own descriptors and load them as plugins during MegaLinter runtime
 
 - Descriptor format is exactly the same than [MegaLinter embedded ones](https://github.com/oxsecurity/megalinter/tree/main/megalinter/descriptors) ([see json schema documentation](https://megalinter.io/json-schemas/descriptor.html))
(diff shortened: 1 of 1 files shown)


wesley-dean:
Screenshot_20250313_095605_edited

This is where the `(raw)` link is added

echoix (approved):

wesley-dean:
@echoix thanks!

echoix:
In remember myself having difficulty finding it, that and the rendered html versions that are hidden at one single place

PR #5099, opened 2025-03-31
Upgrade arm test file version

nvuillam:


echoix (approved):
Only that was missing?

nvuillam:
> Only that was missing?

Yes, it happens every year when the "not older than 2 years version" rule is failing ^^

PR #5257, opened 2025-04-23
Add fmlint (frontmatter linter) to plugins list

wesley-dean:
## Proposed Changes

1. adds a plugin for @adrienverge's [yamllint](https://github.com/adrienverge/yamllint) to lint YAML front matter in Markdown documents.

### Author/Contributor

### Reviewing Maintainer


echoix:
If you could update the title, or include a changelog entry, it would be helpful when doing the release. Otherwise, it's fine if the remaining checks pass

nvuillam:
@wesley-dean please can you update .cspell.json to fix the linter issue ? :)

wesley-dean:
Will do.

wesley-dean:
The typo has been corrected and the cspell list of ignored words has been updated.  Thanks for the feedback!

wesley-dean:
General question: should additions to documentation (e.g., the inclusion of additional plugins) be documented in the changelog?

echoix:
> General question: should additions to documentation (e.g., the inclusion of additional plugins) be documented in the changelog?

Good point. I would go with no. The goal is to have plugins that we don't need to maintain, and are of a different scope of our repo. We happen to also act as the "registry" in the same project, so it may be a bit unfortunate. I believe that plugins should be merged in a timely manner and easier, as they shouldn't be the same level of support as the main project.

echoix (approved):

PR #6809, opened 2025-12-23
feat(linters): change checkmake Docker image reference

bdovaz:
I have changed the image from:

https://hub.docker.com/r/mrtazz/checkmake/tags (which has not been updated since February 2022)

To (runs daily with https://github.com/backplane/conex/blob/main/.github/workflows/docker.yml):

https://hub.docker.com/r/backplane/checkmake/tags

The reason is that it is the only one that is currently being updated and is compatible with arm:

https://hub.docker.com/search?q=checkmake&architecture=arm64

In addition, I have asked in this discussion what the future intention is so that we can rethink the dependency and move to an official one: https://github.com/checkmake/checkmake/discussions/206#discussioncomment-15327252


bdovaz:
@echoix ready for review!

cc @echoix

echoix:
Usually we use official releases, not nightlies, or beta versions.

bdovaz:
> Usually we use official releases, not nightlies, or beta versions.

But did you read the first message? We currently depend on an image tagged as `latest` that hasn't been updated in four years... At least we're going to switch to one in the same condition but that is updated daily.

And besides, look at the comment I posted in the official repository. The idea is that this change will be temporary, but we don't know if “temporary” means a month or a year...

echoix (approved):

PR #6950, opened 2026-01-11
feat: allow selection of platforms to build the custom flavor on

bdovaz:
When merged, https://github.com/oxsecurity/megalinter/pull/2273 will allow someone to create a custom flavor for `linux/arm64`.

There are very few linters that cannot build `linux/arm64`, and the documentation (all_linters.md) lists which ones are compatible and which ones are not.

With this PR, I make it possible for a user to build for `linux/arm64` in addition to building for `linux/amd64`.

This way, we do not block users from this functionality even though we do not yet generate images for `linux/arm64` in our workflows.

This functionality should be considered “beta” since **we are not yet validating or publishing images for this platform in our workflows** so we are unable to detect regressions.

cc @nvuillam @echoix

### DIFF
--- a/.automation/build.py
+++ b/.automation/build.py
@@ -3964,7 +3964,7 @@ def build_custom_flavor(dockerfile):
     work_dir = (
         "/megalinter-builder" if os.path.isdir("/megalinter-builder") else REPO_HOME
     )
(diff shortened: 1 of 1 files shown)


nvuillam:
@bdovaz if it doesnt change the default behavior, all good for me 😃

PR #6978, opened 2026-01-17
Fix .gitignore exclusion list build to include folders starting with a dot

nvuillam:


echoix (commented):
@nvuillam Consider upgrading your locally installed uv , you updated the uv.lock file and it downgraded from a revision 3 to 2 of the schema

PR #6989, opened 2026-01-19
feat: build linters for linux/arm64 where possible in deploy-*-linters.yml

bdovaz:
* Build the **linter_matrix.json** file, also taking into account which linters support `linux/arm64`
* The **deploy-*-linters.yml** workflows use the `ubuntu-24.04-arm` runner to build the linters that support `linux/arm64`
* I have corrected some typos
* I have removed the reference to `multiarch/qemu-user-static:x86_64-aarch64`, which is only necessary when building for ARM from a non-ARM host
* I have disabled (for linux/arm64) several linters, as it is currently not possible to build for ARM with them


bdovaz:
Ready for review @nvuillam @echoix 

For now, we seem to have avoided the 256 limit, so it will be a problem for the future when we add more linters.

nvuillam:
@bdovaz sorry i created you a conflict ^^ (i had to disable LUA_SELENE)

bdovaz:
Resolved

bdovaz:
I don't know why lychee fails on arm runner. When I run it locally with qemu, it works fine...

I also don't know what's happening or why it times out. If I could see more logs...

What do you think? Do you think I can debug it, or should I just make lychee incompatible with arm for now? It's a shame, really, because there shouldn't be any incompatibilities, but that timeout is very strange.

cc @nvuillam @echoix

echoix:
Was it a one-off hang? I launched a rerun of just that. Is it really a failure or just a timeout?

bdovaz:
@echoix looks like it failed again... can you think of how to debug it, if there's anything to debug? The logs aren't very useful, they don't say much...

As a last resort, I would disable ARM compatibility with lychee, but it's a shame because it should work. I don't understand why there is a timeout...

echoix:
Oh, this didn't make it through yet?

bdovaz:
> Oh, this didn't make it through yet?

If you check my previous message I was waiting for a reply to my question from you

bdovaz:
@echoix It looks like the only build that was failing (lychee) has been fixed with https://github.com/oxsecurity/megalinter/pull/7468

So this PR is finally ready to be merged

echoix:
I'll finish reading later

PR #7717, opened 2026-05-07
Fix secretlint issues

bdovaz:
Fix

image


echoix (approved):
Assuming that token never existed or is already useless

PR #7782, opened 2026-05-14
Add file variable

bdovaz:
This change is actually in https://github.com/oxsecurity/megalinter/pull/7150, but to minimize changes to that PR, I'm splitting them into other PRs like this one.

The `grype` and `syft` linters accept files in the `file:{filePath}` format, so I’m adding the `{{FILE}}` variable so that in the descriptor it can be written as `file:{{FILE}}`

Example:

https://github.com/oxsecurity/megalinter/blob/645b813fc3f63108b016f75169fc9b037ddc69bb/megalinter/descriptors/repository.megalinter-descriptor.yml#L460

I've added tests for that method, since I didn't have any before.


bdovaz:
Review please

@echoix @nvuillam

echoix (approved):

PR #7788, opened 2026-05-14
Bump Go Alpine version to 1.26.3-r0 in Dockerfiles

opravil-jan:
## Summary
  
  Alpine edge replaced `go-1.26.2-r0` with `go-1.26.3-r0`, breaking the Docker build with:

  ERROR: unable to select packages:
    go-1.26.3-r0:
      breaks: world[go=1.26.2-r0]
  
  This bumps `GO_ALPINE_VERSION` in `megalinter/descriptors/go.megalinter-descriptor.yml` and regenerates the affected Dockerfiles via `make megalinter-build`.

  ## Affected files

  - `Dockerfile`
  - `flavors/cupcake/Dockerfile`
  - `flavors/go/Dockerfile`
  - `linters/go_revive/Dockerfile`
  - `linters/go_golangci_lint/Dockerfile`
  - `megalinter/descriptors/go.megalinter-descriptor.yml`
  
  ## Test plan

  - [ ] CI Docker build job passes (was failing on `apk add go=1.26.2-r0`)
  - [ ] Affected linter tests (`go_revive`, `go_golangci_lint`) still pass


echoix:
There's nothing wrong with the PR, but I don't understand why we had go pinned so strictly on an apk package. Is there really a version choice on alpine? If we can't really do anything, except having failures, do we really need to pin it?

bdovaz:
@nvuillam will know better because that's where the change was made in this PR:

https://github.com/oxsecurity/megalinter/pull/7577

Actually, the latest version of `go` package in `alpine 3.23` isn't even that, is it? https://pkgs.alpinelinux.org/packages?name=go&branch=v3.23&repo=&arch=x86_64&origin=&flagged=&maintainer=

image

And we have to be careful because when I did a search, I saw that it was changed in that same PR:

https://github.com/oxsecurity/megalinter/blob/6d5e57194d63bde327eadd8365310fbcec6fd8ea/megalinter/descriptors/repository.megalinter-descriptor.yml#L618

nvuillam:
Go needed to be updated, and the newest version was only available on edge packages

If we unpin the version, won't it make the build non-determinists again ?

bdovaz:
> If we unpin the version, won't it make the build non-determinists again ?

When I made the effort of making the installation of all dependencies deterministic, I did the same for APK, but the three of us decided against it because the APK package version is directly tied to the Alpine version, and every time a new version of Alpine is released, we’d have to manually update all the APK dependencies.

That’s what we agreed on, right @echoix ?

echoix:
The next alpine release is due anytime now, until the end of the month.

If I remember correctly, mixing edge and non-edge packages is not a good idea. It just happens to work here.

https://wiki.alpinelinux.org/wiki/Repositories#Edge:


> Warning: Do not use the edge branch in production nor if you desire deterministic, repeatable package installation (such as with containerized environments) via package pinning, as it is under constant development. Instead, use a stable release branch.


Also, further below


> Warning: Do not enable main/community repositories from both the stable release branch and from edge at the same time. This can break your system. Either use edge or stable. If you mix stable and edge repositories, you're on your own.


Alpine releases twice a year, so it's like staying on "LTS" of more constant base packages, like go.

echoix:
> Go needed to be updated, and the newest version was only available on edge packages
> 
> 
> 
> If we unpin the version, won't it make the build non-determinists again ?

Well, you couldn't really rebuild an older version anyways, when they have a new package version, the old one is deleted not long after.

And that version requested doesn't really prevent changing what's returned, it's not a lock file.

Installing from the OSes repo mean trusting the OS'es packagers and the packages that are built.

nvuillam:
@echoix ok so in that case, when alpine is really late according to the version we need (etc GO), better not use alpine packages to install it ?

echoix:
Is this still relevant after upgrading to alpine 3.24?

nvuillam:
I do'nt think so ^^

PR #7790, opened 2026-05-14
Enhance: Optimize linter installations for faster builds

nvuillam:
Switched several linters (LUA_STYLUA, CLOJURE_CLJ_KONDO, KUBERNETES_KUBESCAPE, REPOSITORY_LS_LINT, ENV_DOTENV_LINTER) to use pre-built Alpine-compatible binaries or packages. This reduces build times and improves multi-architecture compatibility by avoiding source compilation, interpreter installs, and reliance on the `sgerrand glibc-compat` layer.


.github/workflows/deploy-DEV.yml
```diff
@@ -127,15 +131,29 @@ jobs:
...
           build-args: ${{ steps.setup.outputs.build-args }}
-          load: false
-          push: false
           secrets: |
             GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }}
           tags: ${{ steps.meta.outputs.tags }}
-          cache-from: type=gha,scope=${{ steps.setup.outputs.cache-scope }}
-          cache-to: type=gha,mode=max,scope=${{ steps.setup.outputs.cache-scope }}
+          cache-from: type=gha,scope=beta-main
+          outputs: type=docker,dest=/tmp/megalinter-image.tar
```
echoix: It isn't possible to compress here? Or it is tested slower?

echoix: https://docs.docker.com/build/exporters/oci-docker/

.github/workflows/deploy-DEV.yml
```diff
@@ -127,15 +131,29 @@ jobs:
...
         timeout-minutes: 90
 
+      - name: Compress image tarball
+        shell: bash
+        run: |
+          set -euo pipefail
+          zstd -3 -T0 --rm /tmp/megalinter-image.tar -o /tmp/megalinter-image.tar.zst
+          ls -lh /tmp/megalinter-image.tar.zst
```
echoix: See previous, if compression would be done by docker when exporting, this serial step might not be needed

.github/workflows/deploy-DEV.yml
```diff
@@ -152,7 +170,6 @@ jobs:
       - name: Checkout Code
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
-          fetch-depth: 0
```
echoix: Isn't that used for some commit history linters?

.github/workflows/deploy-DEV.yml
```diff
@@ -246,7 +256,6 @@ jobs:
...
```
echoix: Again, used to run on git history of our repo?

.github/workflows/deploy-DEV.yml
nvuillam: fetch-depth: 0 is still active on Megalinter workflow , and it's called in every PR, so let's save some time in the build workflow :)

.github/workflows/deploy-DEV.yml
nvuillam: fetch-depth: 0 is still active on Megalinter workflow , and it's called in every PR, so let's save some time in the build workflow :)

.github/workflows/deploy-DEV.yml
nvuillam: Good remark, i try it :)

.github/workflows/deploy-DEV.yml
```diff
@@ -135,21 +135,14 @@ jobs:
             GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }}
           tags: ${{ steps.meta.outputs.tags }}
           cache-from: type=gha,scope=beta-main
-          outputs: type=docker,dest=/tmp/megalinter-image.tar
+          outputs: type=docker,dest=/tmp/megalinter-image.tar,compression=zstd,compression-level=3,force-compression=true
```
echoix, line 139: Shouldn’t it have another extension if it’s tar and compressed? Is it a file that is compressed with zstandard inside an uncompressed tar, or the tar is then compressed by zstandard?

.github/workflows/deploy-DEV.yml
```diff
@@ -135,21 +135,14 @@ jobs:
...
-          set -euo pipefail
-          zstd -3 -T0 --rm /tmp/megalinter-image.tar -o /tmp/megalinter-image.tar.zst
-          ls -lh /tmp/megalinter-image.tar.zst
-
       - name: Upload image artifact
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
         with:
           name: megalinter-image
-          path: /tmp/megalinter-image.tar.zst
+          path: /tmp/megalinter-image.tar
```
echoix, line 164: Possibly keep the previous if the extracted name is changed

flavors/c_cpp/Dockerfile
```diff
@@ -132,7 +132,7 @@ ARG NPM_STYLELINT_CONFIG_STANDARD_VERSION=40.0.0
 ARG NPM_STYLELINT_CONFIG_SASS_GUIDELINES_VERSION=13.0.0
 # renovate: datasource=npm depName=stylelint-scss
 ARG NPM_STYLELINT_SCSS_VERSION=7.0.0
-# renovate: datasource=github-tags depName=dotenv-linter/dotenv-linter
+# renovate: datasource=github-releases depName=dotenv-linter/dotenv-linter extractVersion=^v(?<version>.+)$
```
echoix, line 137: Was there a real need? Renovate’s version coercion is quite good, it’s the whole point of the tool.

megalinter/descriptors/lua.megalinter-descriptor.yml
```diff
@@ -169,26 +164,23 @@ linters:
...
+              apk add --no-cache curl ca-certificates unzip; \
+              case "$TARGETARCH" in \
+                amd64) ARCH=x86_64 ;; \
+                arm64) ARCH=aarch64 ;; \
+                *) echo "unsupported arch: $TARGETARCH" >&2; exit 1 ;; \
+              esac; \
+              curl -fsSL -o /tmp/stylua.zip \
+                "https://github.com/JohnnyMorganz/StyLua/releases/download/v${CARGO_STYLUA_VERSION}/stylua-linux-${ARCH}-musl.zip" && \
+              unzip /tmp/stylua.zip -d /out/bin && \
+              rm /tmp/stylua.zip && \
```
echoix, line 183: I think there might be a caching issue in multiplatform builds. I just remember that we couldn't use shell's "case"/"esac" in a RUN with multi platform. The cache isn't invalidated as the text remains the same

.github/workflows/deploy-DEV.yml
nvuillam: I have no idea, let's see if it works like that !

Note: I use zstd compresson only for internal workflow, not on pushed images

flavors/c_cpp/Dockerfile
nvuillam: as we use executables published with github releases, it seems more logical to watch the releases, because new tags might be pushed without associated releases

echoix: Some of the projects don't really mark them as releases on github. Also, if I remember correctly, github-releases matches the release name, and it is at the will of the release author to include a version name or not, or even to mark the correct one or not.

nvuillam: @echoix it has been verified that releases are handled on the repo before use them as renovate criteria ^^

.github/workflows/deploy-DEV.yml
nvuillam: @echoix tested and working :)

PR #7822, opened 2026-05-15
WIP: Forgejo / Gitea support

### LINKED ISSUE #7779: Forgejo / Gitea support
I'm currently looking into making Megalinter work for Forgejo / Gitea 🙂  (Actions that is)

As they expose basically everything necessary (supposedly) as GHES, the setup is very similar to just Github Actions.

I'm testing with an internal Forgejo system + Forgejo Runner system, with a corporate CA (non-public)
Up to now I tested the workflow including commit fixes, which seems to work!
Couldn't get the reporter to work yet though.

Aka. working on initial docs for that 👍🏻 

Cheers, Lukas

lumarel:
As already mentioned in the issue currenlty looking into making megalinter work for Forgejo / Gitea.
Mostly it seems like it will just work, there are only a few details that make it break.

This is mostly a PoC for now, still have to test if the changes of the reporter actually work,
though if you want you can already give feedback, might just not be finished yet.

Fixes #7779

## Proposed Changes

Currently the plan is to:

1. Add documentation how to setup the workflow
2. Adapt GithubCommentReporter to work with the Quirks of the different platform

### Author/Contributor

### Reviewing Maintainer


docs/install-forgejo-gitea.md
```diff
@@ -0,0 +1,159 @@
...
+
+# Forgejo / Gitea
+
+Use the following Forgejo Actions [YAML template](https://forgejo.org/docs/latest/user/actions/reference)
+
+This CI is heavily based on Github Actions, and due to that will mostly use the same functionality.
+
+## Single Repository
+
+Add the following to an `azure-pipelines.yaml` file within your code repository:
```
nvuillam, line 17: azure-pipelines ? ^^

lumarel: Oops, I took this one as the template and must have overlooked that 🙂

lumarel:
Just a short headsup, it seems like I have to replace nearly all pygithub commands with requests calls, first full draft might still be lots of conditions in the GithubCommentReporter, but in the end it might be better to split it out into it's own reporter.

nvuillam:
@lumarel i agree, if the github libraries are not matching forgejo / gitea apis, let's create new reporters :)

PR #7865, opened 2026-05-19
Move .devcontainer setup from Dockerfile to json file

bdovaz:
* Install Node.js using the [official feature](https://github.com/devcontainers/features/tree/main/src/node).
* Move pip and npm dependencies install to [postCreateCommand](https://containers.dev/implementors/json_reference/).
* Add a new renovate regex for packages parsing in json files (NPM_MARKDOWN_TABLE_FORMATTER_VERSION in this case). The Docker image and its features [are covered by the manager](https://docs.renovatebot.com/modules/manager/devcontainer/).
* Ignore `devcontainer.json` in `jsonlint` because it uses cjson, and jsonlint does not allow to configure modes on a per-file basis. [The specification allows to use cjson](https://containers.dev/implementors/spec/#devcontainerjson), but it must have the `.json` extension.


bdovaz:
Ready @nvuillam @echoix

renovate.json5
```diff
@@ -115,12 +115,14 @@
...
         '# renovate: datasource=(?<datasource>[a-z-.]+?) depName=(?<depName>[^\\s]+?)(?: (lookupName|packageName)=(?<packageName>[^\\s]+?))?(?: versioning=(?<versioning>[^\\s]+?))?(?: extractVersion=(?<extractVersion>[^\\s]+?))?(?: registryUrl=(?<registryUrl>[^\\s]+?))?\\s+(?:ENV|ARG)\\s+[A-Za-z0-9_]+?_VERSION[ =]["\']?(?<currentValue>.+?)["\']?\\s',
         '# renovate: datasource=(?<datasource>[a-z-.]+?) depName=(?<depName>[^\\s]+?)(?: (lookupName|packageName)=(?<packageName>[^\\s]+?))?(?: versioning=(?<versioning>[^\\s]+?))?(?: extractVersion=(?<extractVersion>[^\\s]+?))?(?: registryUrl=(?<registryUrl>[^\\s]+?))?\\\\n(?:ENV|ARG)\\s+[A-Za-z0-9_]+?_VERSION[ =]["\']?(?<currentValue>.+?)["\']?,',
+        '\/\/ renovate: datasource=(?<datasource>[a-z-.]+?) depName=(?<depName>[^\\s]+?)(?: (lookupName|packageName)=(?<packageName>[^\\s]+?))?(?: versioning=(?<versioning>[^\\s]+?))?(?: extractVersion=(?<extractVersion>[^\\s]+?))?(?: registryUrl=(?<registryUrl>[^\\s]+?))?\\s+["\']?[A-Za-z0-9_]+?["\']?:\\s["\']?(?<currentValue>.+?)["\']?\\s',
       ],
     },
```
echoix, line 143: I'm wondering if the renovate jobs timeouts we were seeing often (at least before), were because of the number of lookups and fetching from each url was taking too much time, or if it was the regex having too much to work on. If it would be my second hypothesis, then, adding a third custom manager for only that file type and regex would be appropriate.

Unless the devcontainer json file also needs to check for the other 3 regexes, and all our flavors' action yaml, Dockerfiles, etc also need to match for the newly added regex. (Each file checks for the 4 regexes)

nvuillam (approved):

renovate.json5
bdovaz: Okay, I took the opportunity to do what you suggested and also moved the one I added exclusively for `constants.py` to:  https://github.com/oxsecurity/megalinter/commit/1076931ee336f16c9424e8016175188140f5c7c7#diff-31b1aa2b93afe11792d84c6cb5057a064b5a3294fa0caab92d7570ea604f759f

bdovaz:
@nvuillam What should we do about the osv-scanner error? It has nothing to do with my PR, right?

nvuillam:
@bdovaz i fixed it, if you update your branch with main and merge conflicts it should pass :)

PR #8091, opened 2026-06-15
Make mega-linter-runner/index.js executable in git

Wuodan:
## Fixes

`mega-linter-runner/lib/index.js` is the CLI entrypoint declared in `package.json`, but it was not checked in as executable.

On my Linux box `npm test` thus failed with errors like:

```
/bin/sh: line 1: /home/<user>/.nvm/versions/node/v24.16.0/bin/mega-linter-runner: Permission denied
```

  This PR checks that mode into git and removes the shebang from `mega-linter-runner/lib/cli.js`, which is only imported as a module and is not run directly.

## Proposed Changes

1. Check `mega-linter-runner/lib/index.js` into git with the executable bit set because it is the actual CLI entrypoint.
2. Remove the shebang from `mega-linter-runner/lib/cli.js` because it is used as a module, not as a standalone script.

### Author/Contributor

  - This change is too small to warrant a changelog entry.
  - Not applicable for this change.

### Reviewing Maintainer


nvuillam (approved):
LGTM :)

PR #8093, opened 2026-06-15
Fix mega-linter-runner upgrade CLI test timeout

Wuodan:
The upgrade CLI is interactive and waits for confirmation prompts. The test was invoking `mega-linter-runner --upgrade` without stdin, so it could hang until Mocha timed out.

Pipe confirmation answers into the command so the test runs non-interactively.

## Fixes

`mega-linter-runner --upgrade` is interactive and asks for confirmation before applying changes. The CLI test was running it without any stdin, so it could block waiting for input and eventually fail with a timeout.

This PR makes the test provide the expected confirmation input so it runs non-interactively and reliably.

## Proposed Changes

1. Update the `mega-linter-runner` upgrade CLI test to pipe confirmation input into `mega-linter-runner --upgrade`.
2. Keep the existing assertions unchanged while avoiding the interactive prompt hang.

### Author/Contributor

  - This change is too small to warrant a changelog entry.
  - Not applicable for this change.

### Reviewing Maintainer


nvuillam (changes requested):
@Wuodan to stay compliant with all platforms + coding agents,  I would prefer a new CLI parameter  like `--no-prompt` that would skip the user prompt, plz could you make it this way ?
Thanks :)

Wuodan:
> @Wuodan to stay compliant with all platforms + coding agents, I would prefer a new CLI parameter like `--no-prompt` that would skip the user prompt, plz could you make it this way ? Thanks :)

Done.

This change also runs the --upgrade test in a temp folder because it kept adding a `mega-linter-runner/.gitignore` file which was not know to git.

nvuillam (approved):
LGTM, thanks :)

nvuillam (approved):
Thanks a lot for the PR :)

PR #8132, opened 2026-06-17
Remove unused QEMU from workflows

Wuodan:
## Proposed Changes

1. Removes QEMU setup from workflows which do not use it
2. All those workflows use arm64 runners to build for arm64

### Author/Contributor
  - not needed
  - not needed

### Reviewing Maintainer


nvuillam (approved):
you're right, thanks for the PR :)

PR #8133, opened 2026-06-17
(automation) simplify DEV linter job matrix

Wuodan:
~~Fixes #~~

## Proposed Changes

1. Build the actual `{ linter, platform, runner }` list in `get-linters-matrix` and feed it directly into `strategy.matrix.include`.
2. Drop the `Prepare` step, `job_enabled` no-op jobs, and QEMU setup from the old linter x runner cross-product flow.
3. Use `matrix.platform` directly for build and cache settings, while keeping the existing DEV selection logic and ARM gating.

## Why

The old workflow created many runner/linter combinations that were known up front to do nothing.

It also relied on QEMU for cross-arch builds. With images this large, avoiding unnecessary emulated jobs is worth it on both runtime and CI cost.

Precomputing the real matrix keeps the same behavior with a smaller and cheaper workflow.

## Validation

1. Test run: https://github.com/Wuodan/megalinter/actions/runs/27713374684
2. The workflow was hacked slightly to force ARM64 builds
3. Passed tests for 239/240 images.
4. Built all 240/240 images.
5. Only `swiftlint` on ARM64 failed test cases, which looks isolated to `swiftlint` rather than the matrix change.

### Author/Contributor
  - no changelog entry added
  - no documentation changed

### Reviewing Maintainer


nvuillam (approved):
Thanks a lot :)

PR #8212, opened 2026-06-26
fix(runtime): preserve entrypoint args after user switch

Wuodan:
fix(runtime): preserve entrypoint args after user switch

- install su-exec in the image
- use it in setup-runtime-user instead of su
- keep extra entrypoint arguments intact in the non-root startup path
- remove unused args in action.yml (arguments to entrypoint.sh)

Why:

The extra args in `action.yml` (from 2021) were passed as arguments to the `entrypoint.sh`.
They were always ignored but now caused an error with the previous `su` command for the user-switch.
With the new user-switch with `su-exec` such arguments to entrypoint.sh would be properly handled.

Alternative:

Without su-exec this would also work:
`exec su -p -s /bin/sh "${MEGALINTER_RUNTIME_USER}" -c 'exec /entrypoint.sh "$@"' -- "$@"`
But su-exec is tiny (≈20 kB) and is the dedicated tool for user switching on Alpine.

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -478,6 +478,7 @@ RUN apk -U --no-cache upgrade \
                 make \
                 musl-dev \
                 openssh \
+                su-exec \
                 icu-libs \
                 openjdk21 \
                 lua5.3 \
--- a/flavors/c_cpp/Dockerfile
+++ b/flavors/c_cpp/Dockerfile
@@ -256,6 +256,7 @@ RUN apk -U --no-cache upgrade \
                 make \
                 musl-dev \
                 openssh \
+                su-exec \
                 openjdk21 \
                 py3-pyflakes \
                 cppcheck \
--- a/flavors/ci_light/Dockerfile
+++ b/flavors/ci_light/Dockerfile
@@ -165,6 +165,7 @@ RUN apk -U --no-cache upgrade \
                 make \
                 musl-dev \
                 openssh \
+                su-exec \
                 openjdk17 \
                 libxml2-dev \
                 libxml2-utils \
--- a/flavors/cupcake/Dockerfile
+++ b/flavors/cupcake/Dockerfile
@@ -380,6 +380,7 @@ RUN apk -U --no-cache upgrade \
                 make \
                 musl-dev \
                 openssh \
+                su-exec \
                 openjdk21 \
                 gnupg \
                 php84 \
--- a/flavors/documentation/Dockerfile
+++ b/flavors/documentation/Dockerfile
@@ -254,6 +254,7 @@ RUN apk -U --no-cache upgrade \
                 make \
                 musl-dev \
                 openssh \
+                su-exec \
                 openjdk21 \
                 py3-pyflakes \
                 openjdk17 \
--- a/flavors/dotnet/Dockerfile
+++ b/flavors/dotnet/Dockerfile
@@ -282,6 +282,7 @@ RUN apk -U --no-cache upgrade \
                 make \
                 musl-dev \
                 openssh \
+                su-exec \
                 icu-libs \
                 openjdk21 \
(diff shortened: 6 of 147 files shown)


Wuodan:
In one of my workflows after `megainter the create PR action `peter-evans/create-pull-request` failed with errors like

```
  warning: unable to unlink 'megalinter-reports/updated_sources/expected/targets-unsupported.json': Permission denied
  warning: unable to unlink 'megalinter-reports/updated_sources/expected/workflows-glob-selection.json': Permission denied
  warning: unable to unlink 'megalinter-reports/updated_sources/scripts/render-fixtures.sh': Permission denied
  Switched to branch 'main'
  Your branch is up to date with 'origin/main'.
  /usr/bin/git reset --hard origin/main
  HEAD is now at 361c8cc Add lint workflow using 'oxsecurity/megalinter'
  /usr/bin/git rev-list --right-only --count main...1b789f7a-4b8f-41f2-afe3-90efb292bfdc
  1
  /usr/bin/git -c protocol.version=2 fetch --no-tags --progress --no-recurse-submodules --force --depth=11 origin megalinter/fixes-28172311434:refs/remotes/origin/megalinter/fixes-28172311434
  fatal: couldn't find remote ref megalinter/fixes-28172311434
  Pull request branch 'megalinter/fixes-28172311434' does not exist yet.
  /usr/bin/git checkout --progress -B megalinter/fixes-28172311434 1b789f7a-4b8f-41f2-afe3-90efb292bfdc --
  error: The following untracked working tree files would be overwritten by checkout:
  	megalinter-reports/.cspell.json
  	megalinter-reports/IDE-config.txt
```

so I tried using the new non-root feature (built beta image in my fork) by adding these 3 env-vars:

- MEGALINTER_UID: from `$(id -u)`
- MEGALINTER_GID: from `$(id -g)`
- MEGALINTER_USER: from `$(id -un)`

The extra-args in the `action.yml` caused a problem in the final `exec su` call in `sh/setup-runtime-user`.

Either removing the args or using `su-exec` would fix the issue, this PR does both.

action.yml
```diff
@@ -9,9 +9,6 @@ outputs:
 runs:
   using: "docker"
   image: "docker://ghcr.io/oxsecurity/megalinter:v9.5.0"
-  args:
-    - "-v"
-    - "/var/run/docker.sock:/var/run/docker.sock:rw"
```
echoix: What is this part doing, for what reason it needs to be removed?

Wuodan: those end up being arguments to `entrypoint.sh`. Which currently ignores them and does not pass them to the python code.

But with the new non-root behaviour the `exec su` at `sh/setup-runtime-user:61` threw an error.

This PR fixes both:
- remove the unused args in `action.yml`
- use `su-exec` (the dedicated tool for this on Alpine) instead of `exec su` so if the entrypoint.sh ever expects args it works.

Wuodan:
The PR has 2 commits (with equal message for squashing).

1. the small config changes
2. the large genereated changes from 'make megalinter-build'.

nvuillam:
@Wuodan i merged a big PR that removes docker in docker, please can you handle conflicts ?

Wuodan:
> @Wuodan i merged a big PR that removes docker in docker, please can you handle conflicts ?

Done, thank you. I would still merge the remaining changes as `su-exec` is the better tool to switch user than `su` directly. And if you ever support params to entrypoint to be passed to python code, then the current implementation would fail.

Wuodan:
Sorry, I hit close button by accident.

nvuillam:
I let @echoix or @bdovaz validate, they are better in docker than me 😇

echoix (approved):

PR #8225, opened 2026-06-27
Renovate fix

nvuillam:


echoix (commented):
Let’s see if it’s a good solution

nvuillam:
@echoix only 2 linters depending from repology, and if it's down at least it does not block the whole renovate actions for all others ^^
Seems to work for now :)

PR #8297, opened 2026-07-03
Add ENABLE_DISABLE_LINTERS_PRIORITY to control linter activation precedence

### LINKED ISSUE #8296: Feature: Allow users to select the override mode for ENABLE/DISABLE_LINTERS

I'm using a configuration where I would like to inherit from a shared project configuration using `EXTENDS` and then disable some of the linters from the inherited configuration.

This is not currently possible because `ENABLE` overrides `DISABLE`.

I would like to be able to dictate the override order of `ENABLE_LINTERS` and `DISABLE_LINTERS`.

Something like `ENABLE_DISABLE_LINTERS_PRIORITY: ENABLE|DISABLE` could work (with `ENABLE` being the default to preserve backwards compatibility.

- Creating a meta-language wrapper over the whole configuration
  - ❌ Difficult to maintain and will drift
- Only inheriting from the smallest set of configuration values
  - ❌ Impractical as a best practice

**Additional context**

This was discovered while working on the ability to easily create and publish custom reduced images and run megalinter locally in https://github.com/trevor-vaughan/megalint-config

trevor-vaughan:
Fixes #8296

## Proposed Changes

Add an ENABLE_DISABLE_LINTERS_PRIORITY variable (ENABLE default, DISABLE opt-in) that lets DISABLE_LINTERS override ENABLE_LINTERS for linters present in both lists. Behavior is unchanged when a linter is in at most one list, preserving backward compatibility.

### Author/Contributor

### Reviewing Maintainer


nvuillam (approved):
Thanks for the PR :)

PR #8359, opened 2026-07-09
fix: add SELinux support for volume mounts

MRDGH2821:
Fixes mega-linter-runner on SELinux systems like Fedora.

https://github.com/oxsecurity/megalinter/pull/8214#issuecomment-4816487327

## Proposed Changes

1. Append `z` mount option

### Author/Contributor

### Reviewing Maintainer


echoix:
Is the program "getenforce" available everywhere? What happens when it is not? The mega-linter-runner doesn't necessarily run only on fedora or centos

MRDGH2821:
> Is the program "getenforce" available everywhere? What happens when it is not? The mega-linter-runner doesn't necessarily run only on fedora or centos

Fair point, I added try-catch block for that case. so now this function will evaluate to false on non-SE systems (including windows & mac).

nvuillam:
I would prefer to have a new cli option --se-linux to avoid to try a call getenforce for each call (to not degrade performances on windows / mac / other linux) 
Or maybe there is some ENV variable that could allow to detect we are on SELinux ?

MRDGH2821:
There doesn't seem to be any such env variable. I could find the following approaches:

<details><summary> Calling `sestatus` command </summary>
<p>

```sh
sestatus
```

Output:

```
SELinux status:                 enabled
SELinuxfs mount:                /sys/fs/selinux
SELinux root directory:         /etc/selinux
Loaded policy name:             targeted
Current mode:                   enforcing
Mode from config file:          enforcing
Policy MLS status:              enabled
Policy deny_unknown status:     allowed
Memory protection checking:     actual (secure)
Max kernel policy version:      35
```

</p>
</details> 

<details><summary> Reading `/etc/selinux/config` </summary>
<p>

```sh
cat /etc/selinux/config
```

Output:

```
# This file controls the state of SELinux on the system.
# SELINUX= can take one of these three values:
#     enforcing - SELinux security policy is enforced.
#     permissive - SELinux prints warnings instead of enforcing.
#     disabled - No SELinux policy is loaded.
# See also:
# https://docs.fedoraproject.org/en-US/quick-docs/getting-started-with-selinux/#getting-started-with-selinux-selinux-states-and-modes
#
# NOTE: In earlier Fedora kernel builds, SELINUX=disabled would also
# fully disable SELinux during boot. If you need a system with SELinux
# fully disabled instead of SELinux running with no policy loaded, you
# need to pass selinux=0 to the kernel command line. You can use grubby
# to persistently set the bootloader to boot with selinux=0:
#
#    grubby --update-kernel ALL --args selinux=0
#
# To revert back to SELinux enabled:
#
#    grubby --update-kernel ALL --remove-args selinux
#
SELINUX=enforcing
# SELINUXTYPE= can take one of these three values:
#     targeted - Targeted processes are protected,
#     minimum - Modification of targeted policy. Only selected processes are protected.
#     mls - Multi Level Security protection.
SELINUXTYPE=targeted

```
</p>
</details> 

<details><summary> Calling `getenforce` command </summary>
<p>

```
getenforce
```

Output:

```
Enforcing
```

</p>
</details> 

One thing which could be done is to let the user set `SELINUX` env variable (just like https://12factor.net/config). 
Or as @nvuillam suggested, have cli flag `--se-linux`.

In addition to CLI flag, I also suggest the ability to specify it in megalinter config file:
https://github.com/oxsecurity/megalinter/issues/6601#issuecomment-4817362753

MRDGH2821:
I refactored the logic to read `SELINUX_MODE` env variable. And added a doc as well.

nvuillam (approved):
Works for me, thanks ! :) (i'll merge once we'll have fixed the CI issues)

echoix (approved):

PR #8362, opened 2026-07-09
Honor ADDITIONAL_EXCLUDED_DIRECTORIES in changed-files mode

### LINKED ISSUE #8360: `ADDITIONAL_EXCLUDED_DIRECTORIES` is ignored when `VALIDATE_ALL_CODEBASE: false` (changed-files mode)
**Describe the bug**

`ADDITIONAL_EXCLUDED_DIRECTORIES` is only honored on full-codebase runs. When MegaLinter builds its file list from `git diff` (i.e. `VALIDATE_ALL_CODEBASE: false`, the default in PR/incremental runs), the excluded directories are **not** pruned, so files inside them are linted anyway.

The documentation states that `ADDITIONAL_EXCLUDED_DIRECTORIES` is `List of additional excluded directory basenames. They're excluded at any nested level.` ([`docs/config-variables.md`](https://megalinter.io/latest/config-variables/)). There is no indication that the behavior changes for `VALIDATE_ALL_CODEBASE: false`.

**To Reproduce**

1. Add a change that should be excluded
2. Run `megalinter`

**Expected behavior**

`ADDITIONAL_EXCLUDED_DIRECTORIES` is always honored.

**Additional context**

`MegaLinter.collect_files()` (`megalinter/MegaLinter.py:785`) dispatches:

```python
elif self.validate_all_code_base is False:
    all_files = self.list_files_git_diff()   # changed-files mode
    ...
else:
    all_files = self.list_files_all()        # full-codebase mode
```

- **Full-codebase path** — `list_files_all()` (`MegaLinter.py:961`) walks the tree with `os.walk` and prunes excluded directories:

  ```python
  excluded_directories = self._normalize_excluded_directories(
      utils.get_excluded_directories(self.request_id)   # includes ADDITIONAL_EXCLUDED_DIRECTORIES
  )
  ...
  dirnames[:] = [d for d in dirnames if d not in excluded_directories and ...]
  ```

- **Changed-files path** — `list_files_git_diff()` (`MegaLinter.py:894`) builds the list purely from `git diff --name-only` and **never calls** `get_excluded_directories()`:

  ```python
  for diff_line in diff.splitlines():
      if os.path.isfile(self.workspace + os.path.sep + diff_line) and not os.path.islink(...):
          all_files += [diff_line]
  return all_files
  ```

trevor-vaughan:
In changed-files mode (VALIDATE_ALL_CODEBASE: false), the file list was
built straight from `git diff --name-only` without consulting the excluded
directories, so files under EXCLUDED_DIRECTORIES / ADDITIONAL_EXCLUDED_
DIRECTORIES were still linted. Full-codebase mode prunes them during the
os.walk traversal, creating inconsistent behavior between the two paths.

Prune diff files whose ancestor directory is excluded so both modes behave
identically. The matching rule (a directory matches by its basename at any
nesting level, or by its workspace-relative path) is consolidated into a
single `_is_excluded_dir()` predicate reused by both the git-diff pruning
and the full-codebase os.walk traversal, so the two modes cannot drift.

Closes #8360

Assisted-by: Claude Opus 4.8

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ Note: Can be used with `oxsecurity/megalinter@beta` in your GitHub Action mega-l
   - Keep the Docker Pulls badge in `docs/index.md` in sync by having `docker_stats.py` also update the hardcoded badge total in `.automation/build.py`
   - Fix outdated links in `docs/descriptors/repository_kingfisher.md`
   - Fix `LINTER_RULES_PATH` not being used to resolve config files for linters using `active_only_if_file_found` (e.g. `REPOSITORY_LS_LINT`, `SPELL_PROSELINT`, `SPELL_VALE`), fixes [#8416](https://github.com/oxsecurity/megalinter/issues/8416)
+  - Honor `EXCLUDED_DIRECTORIES` and `ADDITIONAL_EXCLUDED_DIRECTORIES` in changed-files mode (`VALIDATE_ALL_CODEBASE: false`), so files inside excluded directories are pruned from the `git diff` file list the same way they are during full-codebase validation ([#8360](https://github.com/oxsecurity/megalinter/issues/8360))
 
 - Reporters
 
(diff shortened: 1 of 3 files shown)


nvuillam (approved):
Good catch, thanks for the PR :)

PR #8410, opened 2026-07-12
Skip file lint mode test when list_of_files is supported (CI speedup)

nvuillam:
## Proposed Changes

Optimize CI test duration by not running redundant per-file lint mode tests.

1. **Skip `file` CLI lint mode tests when `list_of_files` is also supported.** For any linter whose descriptor declares both `file` and `list_of_files` in `supported_cli_lint_modes`, the `test_success_file_lint_mode` / `test_failure_file_lint_mode` tests are now skipped — `list_of_files` exercises the same code path and covers the same fixtures more cheaply. The skip lives in `LinterTestRoot.run_lint_mode_test`, so it applies to every linter automatically with no per-descriptor changes.

   Behavior summary:
   - `file` only → `file` tests run (unchanged)
   - `file` + `list_of_files` → `file` tests skipped, `list_of_files` runs
   - `list_of_files` only → `file` already skips (unsupported), `list_of_files` runs (unchanged)
   - `project` → always runs when declared (untouched)

2. **Refresh the repo's Claude skills, rules and agents** to document the `supported_cli_lint_modes` descriptor property and the per-mode test methods (`test_success_*_lint_mode` / `test_failure_*_lint_mode`) introduced in #7150, which the docs still described under the old `test_success` / `test_failure` names.

### Author/Contributor

- [x] Add entry to the CHANGELOG listing the change

### Reviewing Maintainer


bdovaz:
@nvuillam Are you sure about this change? I specifically covered them because this class, among others, addresses those cases differently:

https://github.com/oxsecurity/megalinter/blob/main/megalinter/utilstest.py

Search for `linter.cli_lint_mode == "file".

cc @echoix

nvuillam:
@bdovaz the testing CI job added +10 mn to the workflow, i'm trying to save some time by skipping test on only one file when there will be a test for multiple files
I think that when a linter accepts list_of_files , there are very few use cases where people will want to override with "file" cli_lint_mode, so that would be an acceptable thing to not test it

Or maybe i understood wrong ? (which is totally possible ^^ )

PR #8475, opened 2026-07-20
Fix invalid empty SARIF fixes

### LINKED ISSUE #8474: shellcheck SARIF `fixes` have empty `artifactChanges`, so GitHub code-scanning rejects the entire report
**Describe the bug**

MegaLinter's shellcheck SARIF output emits `result.fixes[]` entries whose `artifactChanges` array is empty. The SARIF 2.1.0 schema requires `fix.artifactChanges` to have `minItems: 1`, so GitHub's code-scanning (`github/codeql-action/upload-sarif`) rejects the whole `megalinter-report.sarif`.

Because the upload is all-or-nothing, a single shellcheck fix suggestion (e.g. an SC2292 "Prefer `[[ ]]` over `[ ]`" note) silently blackholes every finding in the report (zizmor, hadolint, golangci-lint, secret scanners, CVEs) from the GitHub Security tab.

For an SC2292 note, the shellcheck SARIF run contains fixes with a `description` but an empty `artifactChanges`:

```json
{
  "ruleId": "2292",
  "level": "note",
  "message": { "text": "Prefer [[ ]] over [ ] for tests in Bash/Ksh/Busybox." },
  "locations": [ /* ... */ ],
  "fixes": [
    { "artifactChanges": [], "description": { "text": "[" } },
    { "artifactChanges": [], "description": { "text": "]" } }
  ]
}
```

Every shellcheck finding that carries a fix suggestion (any rule whose `shellcheck -f json` output includes a `fix.replacements` array, e.g. SC2086, SC2250, SC2292) produces the same empty-`artifactChanges` shape. The converter emits one `fix` object per replacement, copies the replacement text into `fix.description.text`, and leaves `artifactChanges` empty even though the source JSON carries the full replacement region (`line`/`column`/`endColumn`/`replacement`).

**To Reproduce**

1. Lint a repo containing a shell script that triggers a fixable shellcheck rule (e.g. a `[ … ]` test that raises SC2292 under `--enable=require-double-brackets`) with `BASH_SHELLCHECK` enabled and the `SARIF_REPORTER` active.
1. Feed the resulting `megalinter-report.sarif` to `github/codeql-action/upload-sarif` (or validate against the SARIF 2.1.0 schema directly).
1. Upload/validation fails with `artifactChanges does not meet minimum length of 1`:

```
##[error]Unable to upload "…/megalinter-report.sarif" as it is not valid SARIF:
- instance.runs[0].results[1].fixes[0].artifactChanges does not meet minimum length of 1
- instance.runs[0].results[1].fixes[1].artifactChanges does not meet minimum length of 1
- instance.runs[0].results[2].fixes[0].artifactChanges does not meet minimum length of 1
  … (one per shellcheck fix; the whole report is discarded)
```

Schema reference for `fix.artifactChanges`:

```json
{
  "description": "One or more artifact changes that comprise a fix for a result.",
  "type": "array",
  "minItems": 1,
  "uniqueItems": true,
  "items": { "$ref": "#/definitions/artifactChange" }
}
```

**Expected behavior**

The aggregated SARIF should validate against SARIF 2.1.0. Either:

- Populate `artifactChanges` with the real replacement region and inserted text that shellcheck already provides in its `-f json` `fix.replacements` data, or
- Omit the `fix` object entirely when no concrete `artifactChange` can be expressed (a fix with no artifactChanges describes no change and carries no usable information). Omitting `fixes` is valid SARIF (`result.fixes` is optional); an empty `artifactChanges` is not.

**Screenshots**

Not applicable: this is CLI/CI output (the failing `upload-sarif` log is quoted under "To Reproduce").

**Additional context**

Environment:

- MegaLinter image version: 9.6.0
- Descriptor/linter: BASH_SHELLCHECK, shellcheck v0.11.0
- SARIF converter: `shellcheck-sarif` 0.8.0 (Rust `sarif-rs` / `serde_sarif` crate), pinned in `megalinter/descriptors/bash.megalinter-descriptor.yml`
- Reporter: `SARIF_REPORTER` (aggregated `megalinter-report.sarif`)
- Consumer: `github/codeql-action/upload-sarif` (v4.x)

Root cause:

MegaLinter does not convert shellcheck output to SARIF itself. `megalinter/linters/ShellcheckLinter.py` (`manage_sarif_output`) pipes `shellcheck -f json` into the upstream `shellcheck-sarif` binary (`sarif-rs` / `serde_sarif`). That converter writes one `fix` per shellcheck replacement but does not carry the replacement region across, leaving the required `artifactChanges` array empty.

MegaLinter already runs a per-linter SARIF sanitizer, `fix_sarif()` in `megalinter/utils_sarif.py` (invoked via `normalize_sarif_files()` from `Linter.manage_sarif_output`), which corrects other SARIF-validity problems (line-0 regions, missing `version`/`informationUri`, duplicate rule ids, missing `results` array). It does not currently touch `result.fixes`, so the empty `artifactChanges` passes through and the aggregated report is invalid.

Suggested fix:

Normalize `result.fixes` inside `fix_sarif()` so the fix covers every linter rather than shellcheck alone: drop any `fix` whose `artifactChanges` would be empty, and remove the `fixes` key once it becomes empty. This is minimal and information-preserving: it drops only empty fixes and keeps every finding and its location. Equivalent jq transform:

```jq
(.runs[]?.results[]?) |= (
  if has("fixes")
  then (.fixes |= map(select((.artifactChanges // []) | length > 0)))
       | (if (.fixes | length) == 0 then del(.fixes) else . end)
  else .
  end
)
```

Workaround in place downstream:

We post-process `megalinter-report.sarif` before upload with the jq transform above and re-verify no empty `artifactChanges` remain. This is a stopgap in our action pipeline; the correct fix is emitting valid SARIF from MegaLinter so consumers do not each have to sanitize it.

Sanjays2402:
Fixes #8474

## Proposed Changes

1. Remove SARIF fixes whose `artifactChanges` list is missing or empty before aggregation.
2. Preserve valid fixes and remove the `fixes` key when no valid entries remain.
3. Add focused regression coverage for both cases.

### Author/Contributor
- [x] If documentation is needed for this change, has that been included in this pull request (not needed for this internal SARIF normalization fix)

### Reviewing Maintainer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ Note: Can be used with `oxsecurity/megalinter@beta` in your GitHub Action mega-l
 
 - Fixes
 
+  - Remove invalid SARIF fixes with empty `artifactChanges` arrays before report aggregation, fixes [#8474](https://github.com/oxsecurity/megalinter/issues/8474)
   - Write `REPOSITORY_CHECKOV`'s transient GitHub-config scan directory (`branch_protection_rules.json` and similar) to a hidden `.checkov-github-conf` subfolder of the MegaLinter report folder instead of the repository root, so the artifact stays out of the linted tree (gitignored, excluded from file discovery, and skipped by project-mode linters), extending the earlier ansible-lint race-condition fix (#8092)
   - Make remote configuration loading resilient to transient network failures by adding a request timeout and bounded retries with backoff when fetching `MEGALINTER_CONFIG` and `EXTENDS` files over HTTP (fixes intermittent `config_test` failures caused by `raw.githubusercontent.com` CDN cache lag)
   - Disable `TERRAFORM_TERRASCAN` (upstream repo archived by Tenable, unmaintained) and `SQL_TSQLLINT` (no upstream release since 2024-09), as both ship unpatched CVEs with no prospect of a fixed release
--- a/megalinter/tests/test_megalinter/mega_linter_3_sarif_test.py
+++ b/megalinter/tests/test_megalinter/mega_linter_3_sarif_test.py
@@ -9,9 +9,10 @@ import os
 import tempfile
 import unittest
 import uuid
+from types import SimpleNamespace
 from unittest.mock import patch
 
-from megalinter import Linter, MegaLinter, utils_reporter, utilstest
+from megalinter import Linter, MegaLinter, utils_reporter, utils_sarif, utilstest
 from megalinter.constants import DEFAULT_SARIF_REPORT_FILE_NAME
 from megalinter.reporters.SarifReporter import SarifReporter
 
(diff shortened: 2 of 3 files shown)


nvuillam (approved):
Thanks for the fix :)
Will merge once CI agrees ^^

PR #8538, opened 2026-07-26
fix(tests): skip test_api_output on remote-side API endpoint failures

nvuillam:
## Problem

`mega_linter_3_sarif_test::test_api_output` posts to public echo endpoints to verify `ApiReporter` really sends data. It already tries several endpoints in order and skips when none is reachable — but the final "did the endpoints go down?" check **re-probes** them.

That re-probe is racy. An endpoint that returns 503 during the MegaLinter run and answers normally moments later is judged "available", so the test falls through to `self.fail()` — turning a third-party outage into a red build.

On 2026-07-26 an `httpbin.org` outage blocked several unrelated dependency-update PRs at once this way (#8492, #8513, #8517 and others), none of which had anything to do with the API reporter.

## Fix

Decide from what `ApiReporter` actually logged during the run instead of re-probing afterwards. `ApiReporter` already logs failures as `Error posting data to <url> (<status>)`, so the test can classify them:

- every failed post reported a **5xx**, or a connection error occurred → the endpoint failed on its own side → `SkipTest`
- any **4xx** → a genuine payload/configuration problem → still fails

This deliberately keeps real bugs visible: a 400/404, or a mix such as `503 + 400`, still fails the test.

## Verification

The classification was checked against representative reporter output:

| Output | Result |
|---|---|
| `503` on metrics (the observed outage) | skip |
| `503` on logs | skip |
| connection error only | skip |
| `502` + `504` | skip |
| `400` bad request | **fail** |
| `404` not found | **fail** |
| mixed `503` + `400` | **fail** |
| clean success run | no skip |

Test-only change; no linter, descriptor, or user-facing behavior is affected.

### DIFF
--- a/megalinter/tests/test_megalinter/mega_linter_3_sarif_test.py
+++ b/megalinter/tests/test_megalinter/mega_linter_3_sarif_test.py
@@ -7,6 +7,7 @@ Unit tests for Megalinter class
 import glob
 import logging
 import os
+import re
 import tempfile
 import unittest
 import uuid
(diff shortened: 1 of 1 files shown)


echoix:
Would you rather use the new parallel features of GitHub action workflows and launch a local server to respond, that spawns accross the workflow steps? It will need to be allowed through the docker ports though.

nvuillam:
> Would you rather use the new parallel features of GitHub action workflows and launch a local server to respond, that spawns accross the workflow steps? It will need to be allowed through the docker ports though.

This test is not critical enough, your idea is good but i think it would be an overkill , as we just test the capability to call any http endpoint ^^

PR #8606, opened 2026-08-02
Remove deprecated and disabled linters for v10

nvuillam:
## Summary

Prepares the **v10.0.0** major release by removing every deprecated or long-disabled linter, while keeping existing `.mega-linter.yml` files valid.

Removes 14 linters, plus the `API`, `MAKEFILE` and `PUPPET` descriptors which had no other linter left. **MegaLinter no longer provides Makefile and Puppet linting.**

| Linter | Replacement |
| :--- | :--- |
| `API_SPECTRAL` | None — crashing, no upstream fix |
| `JSON_ESLINT_PLUGIN_JSONC` | `JSON_PRETTIER` / `JSON_JSONLINT` |
| `LUA_SELENE` | `LUA_LUACHECK` |
| `MAKEFILE_CHECKMAKE` | None — disabled for security issues |
| `MARKDOWN_REMARK_LINT` | `MARKDOWN_MARKDOWNLINT` |
| `PUPPET_PUPPET_LINT` | None — blocked upstream |
| `REPOSITORY_GITLEAKS` | `REPOSITORY_BETTERLEAKS` |
| `REPOSITORY_KICS` | `REPOSITORY_CHECKOV` |
| `SALESFORCE_LIGHTNING_FLOW_SCANNER` | `SALESFORCE_CODE_ANALYZER_FLOW` |
| `SALESFORCE_SFDX_SCANNER_APEX` / `AURA` / `LWC` | matching `SALESFORCE_CODE_ANALYZER_*` |
| `SQL_TSQLLINT` | None — unmaintained, unpatched CVEs |
| `TERRAFORM_TERRASCAN` | `REPOSITORY_CHECKOV` |

`REPOSITORY_GITLEAKS` is the only one that was still active, so it is the main breaking change for users.

## Backward compatibility

Configuration variables of removed linters are **kept** in the JSON schema and flagged `deprecated: true` with a `(deprecated)` title prefix; their keys stay valid in `ENABLE_LINTERS` / `DISABLE_LINTERS` and `ENABLE` / `DISABLE`. Existing configurations keep validating and are simply ignored at runtime.

MegaLinter now shows a single aggregated notice in the console and PR comments:

> ⚠️ Your configuration references items that have been removed from MegaLinter and are ignored: `API`, `REPOSITORY_GITLEAKS`, `TERRAFORM_TERRASCAN`. See [Removed linters](https://megalinter.io/beta/removed-linters/) to find their replacements.

## New: `megalinter/removed_linters.py`

Single source of truth for every linter and descriptor removed since v6, replacing the `DEPRECATED_LINTERS` list in `build.py` and the hardcoded `["CREDENTIALS", "GIT"]` descriptor list. It drives the schema enums, the deprecation flags, the runtime notice, and a new generated [Removed linters](https://megalinter.io/beta/removed-linters/) documentation page.

Two latent bugs fixed along the way:

- `_is_removed_linter_related_variable()` ran *after* the deprecation-flagging block, so removed variables never received the JSON Schema `deprecated` keyword.
- The old list held `CSS_SCSSLINT`, but the linter actually shipped as `CSS_SCSS_LINT` — the guard never matched its schema properties. The typo is preserved as a schema-enum alias so configurations using it keep validating.

## Secret masking: gitleaks → betterleaks

The internal log secret-masking engine no longer depends on gitleaks at all. `logger.py` now reads a vendored, minified betterleaks ruleset instead of fetching gitleaks' config over HTTP on every run:

- Version pinned in `constants.py`, tracked by a new renovate custom manager and grouped with the betterleaks linter so the two never drift.
- `.automation/vendor_betterleaks_ruleset.py` regenerates the ruleset during documentation builds (nightly auto-update), keeping only rule ids and regexes (66 KB vs 460 KB upstream). It never overwrites the existing file on a fetch failure or a suspicious rule count.
- **Removes a network call from every MegaLinter run**, and redaction coverage nearly doubles: 218 → 408 patterns.

## Tests

67 non-Docker unit tests pass, including a new `removed_linters_test.py`. The `sanitize_string` canary asserting exactly 2 redactions still passes on the larger ruleset, so ordinary linter output is not over-masked.

Docker-based linter tests still need a CI run.

## Notes for reviewers

- `EslintLinter.py` is **not** deleted — it is shared with `JAVASCRIPT_ES_LINT`, `JSX_ESLINT`, `TSX_ESLINT` and `TYPESCRIPT_ES_LINT`.
- Test folders `.automation/test/{lua,salesforce,salesforce-lfs}/` are kept — shared with linters that stay.
- `.cspell.json` is untouched: the removed linter names still appear in the CHANGELOG and on the new documentation page.
- `README.md` linter tables and `docs/descriptors/*` are refreshed by the auto-update workflow, not in this PR.


nvuillam:
cc @echoix @bdovaz :)

echoix:
Can you re-evaluate rumdl? They seem to have evolved a lot since we disabled by default

nvuillam:
@echoix it is still possible to use it by default by setting `MARKDOWN_DEFAULT_STYLE=rumdl` , but i prefer to not let it by default until it becomes the reference markdown linter
For the moment, markdownlint is still the most popular

PR #8682, opened 2026-08-07
docs: state that custom flavors are AGPL-3.0

### LINKED ISSUE #8681: What licence should custom flavor repositories use? Worth a line in the Custom Flavors docs
Hello, and thanks for MegaLinter — the custom flavors feature has cut my CI
image pull time considerably.

While setting one up I hit a question the docs don't currently answer: **what
licence should a custom flavor repository carry?**

MegaLinter is AGPL-3.0. A generated flavor repo contains no MegaLinter source,
but it does contain files produced from the generator's own templates
(`action.yml`, the two workflows, the README), and it publishes a container
image that bundles MegaLinter itself. So it wasn't obvious to me whether:

- the generated scaffolding is considered a derivative work of the AGPL-3.0
  templates,
- publishing the image to a registry counts as conveying AGPL-3.0 software and
  brings the corresponding-source obligation with it, and
- flavor authors are expected to license their repo AGPL-3.0 to match, or are
  free to choose.

Looking at the existing flavor repos on GitHub, there doesn't seem to be a
settled convention: most have no licence file at all (including
`nvuillam/megalinter-custom-flavor-npm-groovy-lint` and
`nvuillam/megalinter-custom-flavor-demo`), and a couple use CC0-1.0. I couldn't
find any using AGPL-3.0.

I've gone with AGPL-3.0 for mine, as the conservative option, and noted in the
README that the published image bundles MegaLinter with a link to upstream
source. But I'd rather follow your intent than guess.

Would you consider adding a short note to the
[Custom Flavors page](https://megalinter.io/latest/custom-flavors/) — even a
single sentence on what licence you expect or recommend for flavor repos? I
suspect most flavor authors haven't thought about it at all, which is arguably
the bigger issue than whichever answer is correct.

Happy to send a docs PR with whatever wording you'd like.

To be clear, I'm not asking for legal advice, just for the project's intent.

laywill:
## Summary

Documents the licence that applies to MegaLinter custom flavors, as confirmed by @nvuillam in #8681:

> MegaLinter custom flavors are still MegaLinter, so are AGPLV3 :)

Fixes #8681

## Changes

**1. `docs/custom-flavors.md` — new "Licensing" section** (what the issue asked for)

States that a custom flavor is still MegaLinter and is therefore covered by AGPL-3.0 — both the generated scaffolding and the published Docker image, which is built from the official MegaLinter image. It then tells the flavor author what to actually do: add an `AGPL-3.0` `LICENSE` file, and keep the link to the MegaLinter source repository in the README.

**2. `mega-linter-runner` custom flavor generator — "License" section in the generated README** (scope expansion, please review on its own merits)

The issue asked only for the docs page, so this second part goes beyond it and I want to be explicit about why I added it rather than slipping it in:

- The point raised in the issue was that *"most flavor authors haven't thought about it at all"*. Someone who never wondered about the licence will not go looking for a Licensing section in the docs — but they do read the README the generator drops in their repository. Putting the statement there is what actually reaches the people the issue is about.
- Surveying existing flavor repos, most carry no licence file at all and a couple use CC0-1.0, i.e. the opposite of the intent confirmed above. A default that states the answer in-place fixes that going forward, whereas a docs-only change only helps authors who already suspected there was a question.
- It keeps the two places consistent: the docs section explicitly says the generated README already contains this statement, so an author has nothing to do beyond adding the `LICENSE` file.

**The two changes are in separate commits**, so if you would rather keep this strictly to the docs page, the second commit can be dropped without touching the first.

## Testing

- Docs-only for commit 1.
- Commit 2 adds static markdown to the EJS template with no new template variables, so generator rendering is unaffected; `test/custom-flavor-generator.test.js` asserts README existence and the `CUSTOM_FLAVOR_LABEL` / repo URL substitutions, none of which this touches. I did not run the runner test suite locally (no `node_modules` in my checkout) — flagging that rather than implying I did.

## Note on wording

I have deliberately kept the wording to a statement of the project's intent plus the two concrete steps, and avoided anything that reads as legal advice. Happy to reword either section however you prefer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -190,6 +190,7 @@ Note: Can be used with `oxsecurity/megalinter@beta` in your GitHub Action mega-l
   - New **`--timeout <seconds>`** option (`-t`): bound the whole run; on expiration the runner prints the container's last log lines, stops and removes the container (no orphan left behind, even on Windows where killing the CLI process does not stop the container), and exits with code 124
   - **Skip `docker pull`** when the requested version is a pinned release tag (`vX.Y.Z`, immutable) and the image is already available locally, removing a useless registry round-trip
   - Print a hint before mounting the workspace when it contains **well-known heavy folders** (build caches, package stores), pointing at `SKIP_CLI_LINT_MODES=project` to keep local runs fast
+  - **`--custom-flavor-setup`**: the generated README now has a **License** section stating the flavor is covered by **AGPL-3.0**, with a link to the MegaLinter source repository
 
 - Dev
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -171,6 +171,7 @@ Note: Can be used with `oxsecurity/megalinter@beta` in your GitHub Action mega-l
   - **megalinter-setup skill**: after install or upgrade, suggest either running MegaLinter locally or creating a pull request and then watching its CI results with megalinter-check
   - **megalinter-check skill**: the first local run now starts with a `--prerun` analysis, so the user can validate the suggested `.mega-linter.yml` performance tuning before the real lint
   - **megalinter-check skill and watcher/runner sub-agents**: the console tips MegaLinter prints (performance warnings, flavor suggestions, `[Activation]` notices, deprecations) are now extracted from the persisted log file and returned to the calling agent in a `tips` field, instead of being lost when reading only the JSON report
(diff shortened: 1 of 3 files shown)


laywill:
> It then tells the flavor author what to actually do: add an AGPL-3.0 LICENSE file, and keep the link to the MegaLinter source repository in the README.

Out of scope of original issue, but could be better if the flavor generator generated an AGPLv3 licence file? Avoids the common issue I saw with custom flavor repos where no licence is specified - ambiguity is worse than minorly irritating someone by overwriting their existing licence?

Candidate follow-on issue perhaps...

nvuillam (approved):
Great, thanks for the PR :)

Issue #8603, opened 2026-08-02
JAVASCRIPT_ES / TYPESCRIPT_ES: document security-plugin usage + pre-bundle a few ESLint security plugins


ofri-peretz:
**What is the current behavior, if applicable?**

The JAVASCRIPT_ES / TYPESCRIPT_ES descriptor documentation lists, among what ESLint detects:

> - **Security Issues**: Detects potential security vulnerabilities (with security plugins)

but no security plugin is named anywhere in the descriptor, and none is bundled in the image. The plugins that *are* pre-installed are `eslint-plugin-import-x`, `-jest`, `-n`, `-prettier`, `-promise`, `-vue`.

So a user who follows that sentence and adds a security plugin to their flat config gets `Cannot find module 'eslint-plugin-...'`, and has to reverse-engineer the `PRE_COMMANDS` recipe out of the error-guidance block in the descriptor. It works, but it's discovered by hitting the error rather than by reading the docs.

Related: the `security` flavor currently bundles 25 linters, none of which analyses JavaScript/TypeScript *source* (the JS-adjacent ones are the Kubernetes manifest linters). Repos on the security flavor get no JS SAST layer at all.

**What is the desired behavior?**

Two parts, and they're independent — the first is useful on its own:

1. **Docs**: a short "Security linting with ESLint" subsection in the JAVASCRIPT_ES / TYPESCRIPT_ES `linter_text`, naming a few plugins and showing a working `PRE_COMMANDS` snippet, so the existing "with security plugins" sentence has somewhere to land.

2. **Bundling**: pre-install a small number of MIT-licensed security plugins next to the existing `eslint-plugin-n` / `-promise` / `-vue`, with renovate-pinned version ARGs following the same pattern, so referencing them in an ESLint config works out of the box.

I have both prepared (descriptor edits for JS + TS, `build.sh` re-run so the generated Dockerfiles and docs stay in sync) and will open the PR if you're open to it.

**Your impression of priority / how important this request is**

 - [ ] Critical: Can't use the tool without it.
 - [ ] Business Critical: Immediate opportunity to win business with this feature
 - [ ] Important: Will significantly enhance the overall utility of the demo
 - [x] Nice to have: self-explanatory

**Business case or other information justifying priority**

It closes a gap the descriptor's own documentation opens, and it's low-risk: no new linter is registered, and nothing changes for users who don't reference these plugins.

**Other information**

[...]

PR #8712, opened 2026-08-10
docs(javascript,typescript): document ESLint security plugin usage

ofri-peretz:
Refs #8603 — the docs half of that issue, on its own.

## The gap

The JAVASCRIPT_ES / TYPESCRIPT_ES feature list includes:

> - **Security Issues**: Detects potential security vulnerabilities (with security plugins)

No security plugin is named in the descriptor, and none is installed in the image — the ESLint plugins bundled alongside are `import-x`, `jest`, `n`, `prettier`, `promise` and `vue`.

So someone who follows that line, adds a security plugin to their flat config and runs MegaLinter gets:

```text
Cannot find module 'eslint-plugin-security'
```

and has to reconstruct the recipe from the `JAVASCRIPT_ES_ERROR_PLUGIN_NOT_FOUND` guidance block. It works — it's just discovered by hitting the error rather than by reading the docs.

## What this adds

A **Security linting with ESLint** section in both descriptors, covering:

- that no security plugin is bundled, and which ones are
- the `PRE_COMMANDS` recipe to install one into the npm root
- the `createRequire` reference, since a bare `import` won't resolve from `/node-deps/node_modules` for the same reason the preceding flat-config section describes
- a short table of commonly used plugins
- a pointer back to workaround 2 for projects that already carry a security plugin in their own `devDependencies`

Documentation only. No linter is registered, no package is added to the image, and nothing changes for users who don't reference these plugins.

## Disclosure

I maintain two of the five plugins listed (`eslint-plugin-secure-coding`, `eslint-plugin-node-security`). They're last in the table, after `eslint-plugin-security`, `eslint-plugin-no-unsanitized` and `@microsoft/eslint-plugin-sdl`, and all five are MIT.

If you'd rather the table named only plugins you have no vendor relationship with, say so and I'll drop mine — the docs gap is worth closing either way, and that's not the part of this I care about.

The bundling half of #8603 is deliberately **not** here. It's the part that would need your judgement on image size and single-vendor risk, and it shouldn't hold up a docs fix.

## Test plan

- `bash build.sh --doc` run to regenerate `docs/descriptors/javascript_eslint.md` and `docs/descriptors/typescript_eslint.md` from the descriptors; both generated files are in the diff.
- The same run also touched `docs/index.md` (Docker pull badge), `docs/articles.md`, `docs/descriptors/python_pylint.md`, `.automation/generated/linter-licenses.json` and two new `docs/licenses/*.md` — all live-fetch drift unrelated to this change, so they're reverted and not in the diff.
- `markdownlint-cli2 --config .github/linters/.markdownlint.json` on the three changed markdown files: the same 8 pre-existing findings before and after, none introduced.
- CHANGELOG.md updated under Unreleased → Doc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ Note: Can be used with `oxsecurity/megalinter@beta` in your GitHub Action mega-l
 - Doc
   - New **Docker pulls per month** graph, showing the growth of MegaLinter adoption since October 2020, displayed in the README and on the [Flavors statistics](https://megalinter.io/latest/flavors-stats/) page
   - Refreshed the **MegaLinter references in linters documentation** (`linter_megalinter_ref_url`): verified all existing links, updated moved pages (ktlint, robocop, csharpier, zizmor, ruff, proselint), and opened 47 suggestion PRs on linters repositories that did not mention MegaLinter yet
+  - New **Security linting with ESLint** section in the JAVASCRIPT_ES and TYPESCRIPT_ES documentation: states that no security plugin is bundled, shows the `PRE_COMMANDS` recipe and the `createRequire` reference needed under flat config, and lists commonly used plugins. Closes the gap left by the "Security Issues (with security plugins)" line, which previously named no plugin and had no working example
 
 - mega-linter-runner
 
--- a/docs/descriptors/javascript_eslint.md
+++ b/docs/descriptors/javascript_eslint.md
@@ -92,6 +92,50 @@ JAVASCRIPT_ES_CLI_EXECUTABLE: node_modules/.bin/eslint
 
 To add a single package with pnpm instead of installing the full tree, run `corepack enable pnpm` first, then `pnpm add --ignore-workspace --no-save <package>`.
 
+### Security linting with ESLint
+
+The feature list above refers to security analysis "with security plugins". MegaLinter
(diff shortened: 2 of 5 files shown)


nvuillam:
@ofri-peretz these doc markdown pages are generated from linter descriptors, so you must directly update linter descriptors yml, then they will be used by auto-update job to regenerate the markdown 😊

nvuillam:
@ofri-peretz i watched better (from a computer, not my phone this time ^^)

I'm ok with the updates, but please can you remove your plugins ? we can't recommend tools that are not already heavily adopted, and unfortunately yours are not (yet !)

ofri-peretz:
Done — mine are out, the table now lists `eslint-plugin-security`, `eslint-plugin-no-unsanitized` and `@microsoft/eslint-plugin-sdl`. That's the right call and no argument from me.

On the first point: the descriptor YAMLs are edited in this PR (`megalinter/descriptors/javascript.megalinter-descriptor.yml` and `typescript...`), and the two `docs/descriptors/*_eslint.md` files in the diff are the output of `bash build.sh --doc`, not hand edits. Happy to drop the generated pair from the branch if you'd rather the auto-update job produce them.

I also merged your `main` merge back in so the branch is current.

nvuillam (approved):
Thanks, will merge once ci is ok :)

PR #8715, opened 2026-08-10
Add SARIF output support for 13 linters + 4 Salesforce Code Analyzer engines

nvuillam:
## Summary

Wires native SARIF output support into linters that already have it upstream but weren't hooked into MegaLinter's `SARIF_REPORTER` yet:

- **zizmor** (`--format=sarif`, stdout) — also recovers a findings-reflecting return code, since `--format=sarif` always exits 0
- **bicep_linter** (`--stdout --diagnostics-format Sarif`)
- **cppcheck** (`--output-format=sarif`, both C and CPP descriptors via the shared linter file)
- **clj-kondo** (`--config '{:output {:format :sarif}}'`)
- **htmlhint** (`--format sarif`, picked up from `htmlhint.sarif` which it writes to the CWD)
- **protolint** (`--add-reporter sarif:<path>`)
- **sqlfluff** (`--format sarif`)
- **swiftlint** (`--reporter sarif --output <path>`)
- **osv-scanner** — re-enabled (`can_output_sarif` was force-disabled with a "disabled until it works!" comment; no blocking bug found in git history)
- **jscpd** — reporters string handled in `JsCpdLinter.build_lint_command()` since jscpd's reporter list is one comma-joined CLI value; picked up from `copy-paste/jscpd-report.sarif` (the actual filename jscpd 5.0.14 writes)
- **lintr** — `RLinter.py` builds its own `R -e` command rather than going through the standard command pipeline, so `lintr::sarif_output()` is invoked directly when SARIF is requested (requires the `jsonlite` R package, now installed, and a manually-set `path` attribute on the lint results — see below)

Also adds SARIF output to the 4 Salesforce Code Analyzer engines (`SALESFORCE_CODE_ANALYZER_APEX/AURA/LWC/FLOW`) via a new shared `SalesforceCodeAnalyzerLinter` class: the `sf code-analyzer` CLI infers its report format from the `--output-file` extension, which is already baked into a fixed argument list, so a second `--output-file` from the standard SARIF injection would collide. The class swaps the extension from `.csv` to `.sarif` only when SARIF output is requested, so normal runs keep producing CSV as before.

**Two linters were dropped from this batch after CI proved the SARIF support isn't actually available in the pinned/released tool version, despite what the initial research found**:
- **roslynator**: the pinned `roslynator.dotnet.cli` 0.13.0 rejects `--output-format sarif` at runtime ("Unknown output format 'sarif'"). The 0.13.0 version bump is kept (works fine otherwise).
- **trufflehog**: `--sarif` was merged upstream on 2026-08-05, *after* the latest release v3.96.0 (2026-07-24) that MegaLinter pins — no released version has it yet.

## Known caveats worth a second look

- `clj-kondo`'s upstream SARIF output currently nests `region` one level too deep under `artifactLocation` (clj-kondo/clj-kondo#2345), and enabling SARIF output together with `EXCLUDED_DIRECTORIES` forwarding in project mode causes the exclude-forwarding to be silently skipped (both share the `--config` flag guard in `CljKondoLinter.manage_excluded_directories_config()`).
- `bicep_linter`'s `--stdout` mixes the compiled ARM template JSON and the SARIF diagnostics in the same stream; `Linter.manage_sarif_output()`'s JSON-block extraction correctly picks out the SARIF-shaped block in CI, but it's a less common pattern worth extra scrutiny if it's ever touched again.
- `osv-scanner`, `jscpd`, and the Salesforce `aura` engine report all (or nearly all) of their SARIF findings at "warning" level rather than "error" — added to the existing `betterleaks`-style exception list in the shared SARIF test rather than the plain `errors > 1` assertion.
- `lintr::sarif_output()` only works on results from `lint_dir()`/`lint_package()` in principle (it reads an internal `path` attribute those set on the lint results, and aborts otherwise); `RLinter.py` sets that attribute manually since MegaLinter calls plain `lint()` per file. The attribute only feeds the SARIF `ROOTPATH` URI, not file I/O, so this is safe but is relying on an internal implementation detail rather than a documented API.

## Test plan

- [x] `test_report_sarif` passes for all linters/engines above — verified in CI across 8 rounds of real bug fixes surfaced by the new tests (a crash in `RoslynatorLinter`'s restore step, wrong filenames/flag placement for jscpd/trufflehog/htmlhint, a missing R package, a forgotten `make megalinter-build`, zizmor's always-0 exit code, a test-fixture activation gap for sqlfluff, several rounds on lintr's SARIF path handling, and two linters — roslynator, trufflehog — whose upstream SARIF support turned out not to be released yet)
- [x] Existing success/failure lint-mode tests still pass


bdovaz:
@nvuillam @echoix 

I see that in utilstest.py you're setting it up so that several linters report everything as a warning instead of warnings and errors, and you're preventing an assert from being triggered.

Back in the day, in several pull requests I made to improve Sarif-related issues, what I did was identify all of that and create issues in their respective official repositories, and even create pull requests where I saw that it was possible.

Because many have warning or error rules, but when you use Sarif, everything is treated as a warning—in other words, support for Sarif is partial.

nvuillam:
@bdovaz you're right, i just launched a task to precise in doc when sarif is partially supported and submit PRs to related repos !

nvuillam:
Good point, thanks. I looked into each of the three linters this touched (osv-scanner, jscpd, the Salesforce Code Analyzer aura engine) plus the pre-existing betterleaks case:

- **osv-scanner**: genuine gap, the CVSS score was already being computed for the `security-severity` property a few lines above where `level` gets hardcoded — opened [google/osv-scanner#2982](https://github.com/google/osv-scanner/pull/2982) mapping Critical/High to `error`.
- **jscpd**: also a real gap (every clone is `warning`, size/token count is available but unused), but there's no existing size cutoff in the codebase to base a fix on, so I didn't want to just invent one — opened [kucherenko/jscpd#908](https://github.com/kucherenko/jscpd/issues/908) proposing it and asking what threshold they'd want.
- **Salesforce Code Analyzer (aura engine)**: turned out not to be a mapping bug — the severity→SARIF-level function is correct (Critical/High → error), the Aura ruleset does have High-severity rules available, our specific bad-fixture just didn't happen to trip one of them. So this one stays excluded from the strict test for now, but I didn't file anything upstream since there's no actual bug to report.

Documented the partial support (with the links above) directly in each linter's `linter_text` so it shows up on the docs pages, in c2266084e4.

PR #8781, opened 2026-08-22
Point at the SARIF report instead of "No output available" in reporters

### LINKED ISSUE #8730: Improve the output of reporters when a linter outputs in SARIF format
When a Linter outputs in SARIF format, “No output available” text shows in the Azure Reporter (and I assume in the other tools as well, since the logic is the same).

https://github.com/oxsecurity/megalinter/blob/7b5f617d0111c4479c934b7630a52a09135df8a9/megalinter/utils_reporter.py#L808

MegaLinter should at least tell you to download the sarif artifact from that linter.

What do you think @nvuillam?

I'm thinking of replacing:

```python
else:
    linter_output = "No output available"
```

With:

> The condition comes from:

https://github.com/oxsecurity/megalinter/blob/7b5f617d0111c4479c934b7630a52a09135df8a9/megalinter/reporters/SarifReporter.py#L61

```python
elif linter.sarif_output_file is not None and os.path.isfile(linter.sarif_output_file):
    linter_output = f"Download the artifact named {linter.sarif_output_file} to check for issues"
else:
    linter_output = "No output available"
```

I'm not sure if this is the right place to put this code, or if it should go somewhere else in this `_build_sections_content` function.

nvuillam:
Closes #8730.

## Problem

A linter reporting in SARIF writes nothing to stdout, so its per-linter details section in every reporter was a dead end:

```
No output available
```

## Fix

The details section now names the SARIF report to open and links the artifacts:

| Case | Rendered |
|---|---|
| SARIF linter, run URL known | `No text output: ruff reports in SARIF format. Download [MegaLinter artifacts](…) and open `sarif/PYTHON_RUFF.sarif`` |
| SARIF linter, no run URL | `No text output: ruff reports in SARIF format. See `sarif/PYTHON_RUFF.sarif` in MegaLinter artifacts` |
| Non-SARIF linter | unchanged |

One helper in `utils_reporter`, used by **both** dead ends — `No output available` and `Linter output file not found` — so the **GitHub**, **GitLab**, **Azure** and **Bitbucket** comment reporters and the markdown summary all benefit from a single change, as @bdovaz expected in the issue.

## Three refinements on the issue's proposal

The issue suggested `f"Download the artifact named {linter.sarif_output_file}"` guarded by `os.path.isfile(...)`. Three things came out of checking that against the code:

**1. `sarif_output_file` is not an artifact name.** It is built at `Linter.py:1959` as `{report_folder}/sarif/{LINTER_NAME}.sarif` — an absolute *container* path like `/tmp/lint/megalinter-reports/sarif/PYTHON_RUFF.sarif`. The artifact is named in the user's own `upload-artifact` step (`MegaLinter reports` in the shipped workflow), so MegaLinter cannot know it. The message names the path **relative to the report folder** instead, and a test asserts the absolute path never leaks into it.

**2. `os.path.isfile()` would have been flaky.** `SarifReporter.processing_order = -9999` — it runs *before* the comment reporters, and at `SarifReporter.py:94` it **deletes** each per-linter `.sarif` when `LOG_FILE: none`. A stat at comment-build time would be `False` in exactly that configuration and the hint would silently vanish. The condition is the linter's SARIF intent (`can_output_sarif` + `output_sarif` + `sarif_output_file is not None`) instead.

**3. `Linter output file not found` is the sibling dead end.** It fires when `TEXT_REPORTER` is off or report files cannot be written; a SARIF linter reaches it too, and the same guidance applies.

## On the artifacts link

No new URL is invented. `build_markdown_summary_footer` already renders this exact value as *"See detailed reports in [MegaLinter artifacts](…)"*, and `action_run_url` was **already** a parameter of `_build_sections_content` — so there is no signature change and the link inherits whatever correctness the footer already has, including the `REPORTERS_ACTION_RUN_URL` override.

On Azure it is already a true deep link: the reporter passes `get_artifacts_url()` (the published-artifacts view) by default.

I deliberately did **not** build platform-specific artifact routes such as `{CI_JOB_URL}/artifacts/browse` or a GitHub `#artifacts` anchor: whether artifacts exist at all depends on the user's own CI config, so those would 404 for anyone without an upload step. When no URL is available (the markdown summary passes `""`), the message degrades to naming the file, mirroring the footer's own fallback.

## Testing

9 new cases in `utils_reporter_test.py` (25 pass in the file), covering: the linked and unlinked renderings, both dead-end branches, non-SARIF linters staying byte-identical, SARIF-capable-but-disabled, SARIF-enabled-without-output-file, and no absolute path leaking.

One robustness fix found while testing rather than assumed: `report_folder` can legitimately be `""` (`Linter.py:346-347`), and `os.path.relpath` against it produced `../../tmp/lint/…`. The path is now derived the same way `get_sarif_arguments()` builds it, which cannot walk out of the report folder — with a regression test.

No descriptor touched, so no `make megalinter-build`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,9 @@ Note: Can be used with `oxsecurity/megalinter@beta` in your GitHub Action mega-l
   - The **API reporter** variables (`API_REPORTER`, `API_REPORTER_URL`…) are not flagged as **deprecated** anymore in the configuration JSON schema: they were collateral damage of the removal of the `API` descriptor in v10.0.0, and IDEs displayed them as obsolete
 
 - Reporters
+  - Linters reporting in **SARIF** format no longer show **No output available** in Pull Request comments and summaries: the details section now names the SARIF report to open and links the **MegaLinter artifacts** ([#8730](https://github.com/oxsecurity/megalinter/issues/8730))
+    - Applies to the **GitHub**, **GitLab**, **Azure** and **Bitbucket** comment reporters and to the markdown summary
+    - The link points where the reporter already links its detailed reports, so it follows `REPORTERS_ACTION_RUN_URL` when you set it
 
 - Flavors
 
--- a/megalinter/tests/test_megalinter/utils_reporter_test.py
+++ b/megalinter/tests/test_megalinter/utils_reporter_test.py
@@ -7,6 +7,7 @@ Unit tests for utils_reporter
 import unittest
 
 from megalinter.utils_reporter import (
+    _build_missing_output_message,
     _build_table_content,
     _sort_linters_by_icon_severity,
     get_linter_status_icon,
@@ -24,7 +25,15 @@ class FakeLinter:
         total_number_warnings=0,
         disable_errors_if_less_than=None,
         linter_name="fake_linter",
+        can_output_sarif=False,
+        output_sarif=False,
+        sarif_output_file=None,
+        report_folder="/tmp/lint/megalinter-reports",
     ):
+        self.can_output_sarif = can_output_sarif
+        self.output_sarif = output_sarif
+        self.sarif_output_file = sarif_output_file
+        self.report_folder = report_folder
         self.status = status
         self.return_code = return_code
         self.number_errors = number_errors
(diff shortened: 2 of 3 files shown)


nvuillam:
@bdovaz does it seem ok for you to fix the issue you declared ? :)

bdovaz:
I think that's a great solution! thanks!

nvuillam:
@bdovaz thanks for your feedback, merged ;)
will be in beta in a few minutes if u wanna test ^^

PR #8794, opened 2026-08-24
Add tofu validate (OpenTofu) as TERRAFORM_TOFU_VALIDATE

### LINKED ISSUE #8793: Add tofu validate command as linter

The `terraform` descriptor currently ships four linters — `TERRAFORM_TFLINT`,
`TERRAFORM_TERRAGRUNT`, `TERRAFORM_TERRAFORM_FMT` and the recently added
`TERRAFORM_TOFU_FMT` — and none of them performs semantic validation of the
configuration. `fmt` only checks canonical formatting, and `tflint` focuses on
provider-specific rules and naming conventions. Neither reports an invalid
resource argument, a wrong attribute type, an undeclared variable, or a broken
module input contract.

`tofu validate` fills that gap: it checks that the configuration is internally
consistent and semantically valid against the actual provider schemas, and it
does so without contacting the remote state or any cloud API.

Add `TERRAFORM_TOFU_VALIDATE`, running `tofu validate` per root module.

One prerequisite drives the whole design: `tofu validate` cannot run on an
uninitialized directory. It needs the provider schemas and the child modules to
be installed, and otherwise fails with `Missing required provider` or
`Module not installed`. The linter therefore has to initialize first:

    tofu init -backend=false -input=false -no-color -lockfile=readonly
    tofu validate -no-color

`-backend=false` is the important flag. It skips backend initialization
entirely, so no remote state is read, no state lock is acquired and no cloud
credentials are needed. That keeps the linter usable on reusable module
repositories that have no backend block at all, and in CI jobs with no cloud
access.

Suggested behaviour:

- **Scope**: `tofu validate` operates on a directory, not on a list of files, so
  `cli_lint_mode: project` (as `TERRAFORM_TFLINT` already does) or one
  invocation per directory containing `.tf` / `.tofu` files.
- **Workspace side effects**: `init` writes `.terraform/` and may rewrite
  `.terraform.lock.hcl`. Pointing `TF_DATA_DIR` outside the workspace keeps the
  repository clean, and `-lockfile=readonly` turns an out-of-sync lock file into
  a reported failure instead of a silent mutation.
- **Non-interactive**: `-input=false` and `TF_IN_AUTOMATION=1` to avoid prompts
  and reduce output noise.
- **Network access**: `validate` itself is offline, but `init` does need network
  access to resolve providers and remote module sources. This is the same
  constraint `tflint --init` already has in this descriptor, handled there with
  `TERRAFORM_TFLINT_SECURED_ENV` and `PAT_GITHUB_COM`; an equivalent pair of
  variables would cover private registries and private module sources.
- **Structured output**: `tofu validate -json` emits machine-readable
  diagnostics with severity, file and line, which would feed error/warning
  counting and SARIF reporting more reliably than a regex over the text output.
- **Configurability**: a `TERRAFORM_TOFU_VALIDATE_INIT_ARGS` variable, plus a
  way to skip the init step for projects that initialize the directory
  themselves beforehand.

MegaLinter already has a per-linter `pre_commands` mechanism in `Linter.py`
(configurable as `<LINTER>_PRE_COMMANDS`) that no descriptor currently uses
declaratively. It may be the least invasive place to attach the `init` step.

- Running `tofu validate` without `init`: not viable, it errors out on any
  configuration that declares providers or modules.
- `tofu init -backend=false` as a standalone linter: it does verify dependency
  resolution integrity (unresolvable module sources, unsatisfiable version
  constraints, lock file drift), but it produces no diagnostics of its own, so
  it is more valuable as the prerequisite of `validate` than as a linter.
- `terraform validate`: equivalent functionality, but OpenTofu is MPL-2.0 and
  the `tofu` binary is already in the image.

**Additional context**

The `tofu` binary has already been added from: https://github.com/oxsecurity/megalinter/issues/8729

eloyred:
Closes #8793.

Adds **`tofu validate`**, the built-in validator of [OpenTofu](https://opentofu.org/), as **`TERRAFORM_TOFU_VALIDATE`**.

It fills a real gap in the `terraform` descriptor: `tflint`, `terragrunt`, `terraform fmt` and `tofu fmt` all skip semantic validation, so nothing currently reports an unsupported argument, a wrong attribute type, a reference to an undeclared variable or local, or a broken module input contract.

## Initialization is a hard prerequisite

`tofu validate` cannot run on an uninitialized directory: it needs the provider schemas and the installed child modules. `TofuValidateLinter` therefore runs, once per analyzed directory:

```shell
tofu -chdir=<directory> init -backend=false -input=false -no-color
```

`-backend=false` is what keeps this a linter rather than an infrastructure operation: backend initialization is skipped entirely, so no state is read, no state lock is acquired and no cloud credentials are needed. It works on reusable module repositories that declare no backend at all, and in CI jobs with no access to the infrastructure.

The initialization is cached per directory, so a module with 10 `.tf` files is initialized once and validated ten times, not initialized ten times. An initialization failure is returned as the linter error, because an unresolvable provider or module source is a genuine finding, not an environment glitch.

Init arguments are overridable through **`TERRAFORM_TOFU_VALIDATE_INIT_ARGUMENTS`**, documented on the linter page. Adding `-lockfile=readonly` turns an out-of-sync `.terraform.lock.hcl` into a reported error instead of letting the initialization update it. It is deliberately **not** the default: a repository with no committed lock file would then fail on every run.

## Lint mode

**`list_of_files`**. `tofu validate` accepts no path argument at all — the module is selected by the global `-chdir` option, which OpenTofu only accepts *before* the subcommand. A single run collects the directories of the analyzed files and initializes and validates each one exactly once.

`file` mode was my first attempt and it is wrong here, because it repeats a module's diagnostics for every file that module contains. Measured on a real OpenTofu module repository: one undeclared-variable typo was reported as **6 errors across 6 files**, 5 of them innocent, with `validate` running 6 times. The same typo in `list_of_files` mode reports **2 errors** — one per module genuinely broken, as that repository's `examples/docs` instantiates the root module — with 2 validate runs, and no file wrongly flagged.

`project` mode is deliberately **not** declared: `tofu validate` has no recursive option, so a single invocation at the workspace root would validate the root module only and silently ignore every module in a subdirectory. Verified on a 50-module repository whose root holds a single `versions.tf`: a root-only `init -backend=false` + `validate` reports `Success! The configuration is valid.` having validated no resource, module or variable at all.

## File extensions

**`.tofu`** only, the same scoping `TERRAFORM_TOFU_FMT` uses, so `.tf` stays free for a future `terraform validate` linter and enabling both would not validate the same repository twice. Repositories keeping their OpenTofu configuration in `.tf` files opt in, as documented on the linter page:

```yaml
TERRAFORM_TOFU_VALIDATE_FILE_EXTENSIONS: [".tofu", ".tf"]
```

Two asymmetries with `tofu fmt` are worth flagging, both documented in `linter_text`.

**The extensions select the directories to validate, not the files to report on.** `tofu validate` always validates a whole module, so once a directory has been selected through a `.tofu` file, every `.tf` file in that same directory is parsed too and can produce diagnostics. Verified — a directory holding `main.tofu` plus `other.tf` reports diagnostics from both. So the extension split deconflicts the pure-`.tf` and pure-`.tofu` cases, not mixed directories.

**`.tfvars` is deliberately not offered**, although `tofu fmt` does format those files. `tofu validate` ignores variable definition files entirely: an undeclared variable in a `test.auto.tfvars` yields the `Value for undeclared variable` warning under `plan`, but `validate` reports nothing. And a directory holding only `.tfvars` files is an empty directory for OpenTofu — `init` says `OpenTofu initialized in an empty directory!` and `validate` returns `Success! The configuration is valid.` with exit 0. Allowing the extension would run a pointless initialization in every variables folder and report a misleading success.

## Install

No new install: it reuses the existing `ghcr.io/opentofu/opentofu:${TERRAFORM_TOFU_FMT_VERSION}` stage and `/usr/bin/tofu`, the same way `TERRAFORM_TERRAFORM_FMT` reuses `TERRAFORM_TERRAGRUNT_VERSION`. The build deduplicates the `FROM` / `COPY` lines in the main and flavor Dockerfiles, and emits them standalone in `linters/terraform_tofu_validate/Dockerfile`.

## Fixtures

Both fixtures declare **no provider and no module**, so the initialization resolves nothing and needs no network access: the tests stay hermetic and fast, and no `.terraform` directory is written into the analyzed sources.

They live in `good/` and `bad/` subfolders because `tofu validate` validates a whole directory: a bad file sitting next to a good one would fail the success test.

## Warnings are not counted

`cli_lint_errors_count: regex_count` on `Error: ` only — no `cli_lint_warnings_count`. In OpenTofu 1.12 the old configuration deprecations that used to be warnings are hard **errors** (quoted type constraints), or have been dropped entirely (interpolation-only expressions). The warnings that remain are provider-driven (`Deprecated Attribute`…), so exercising a warning counter would require a fixture that downloads a real provider, and `test_linter_failure` rightly requires a declared counter to be exercised by the fixture. Warnings still show up in the linter output, they are just not counted separately.

## Self-contained repositories

Initialization resolves dependencies, so a `module` block whose local `source` escapes the repository root (`../../modules/something`) can not be installed and surfaces as `Unreadable module directory`. There is a new `common_linter_errors` entry and a `linter_text` paragraph for it, because I hit it on a real repository: a template repository whose 50 directories only become valid modules after a generator symlinks them into another tree. Any tool resolving module sources fails there the same way — the limitation is the repository not being self-contained, not the linter.

## Verification

Every fact was checked against the real tool rather than taken from the documentation.

| Checked | Result |
|---|---|
| `tofu --version` | `OpenTofu v1.12.6` → feeds `version_extract_regex`, reported as `tofu-validate v1.12.6` |
| Init command built | `tofu -chdir=. init -backend=false -input=false -no-color` → exit 0, no network access |
| Validate command built | `tofu -chdir=. validate -no-color` |
| Good fixture | `Success! The configuration is valid.`, exit 0 |
| Bad fixture | exit 1, two diagnostics: undeclared input variable + undeclared local value |
| Error count | both diagnostics prefixed `Error: ` → `regex_count` = 2 |
| Artifacts | no `.terraform` nor `.terraform.lock.hcl` written for provider-less modules |
| Fixture formatting | `tofu fmt -check -diff -recursive` clean on both fixtures |
| `.tofu` support | a module made only of `.tofu` files initializes and validates normally |
| Directory scoping | a directory holding `main.tofu` + `other.tf` reports diagnostics from both files |

Beyond the fixtures, the linter was run against three real OpenTofu repositories, which is where the lint mode problem and two of the `common_linter_errors` entries come from:

| Repository shape | Outcome |
|---|---|
| Self-contained module, real `hashicorp/aws` provider, no committed lock file, 2 module directories | 6 files, 0 errors, init created the lock file, 2 inits and 2 validates |
| Same repository with one undeclared-variable typo injected | 2 errors in `list_of_files` mode, against 6 errors over 6 files in `file` mode |
| 50-directory template repository whose module sources escape the repository root | `Unreadable module directory`, and a root-only run reports success having validated nothing |
| Module repository with private `ssh://git@github.com` module sources | `Failed to download module` / `Host key verification failed`, no credentials in the container |


nvuillam:
Thanks @eloyred :)

Question: how does tofu-validate act on .tf files that are handled by terraform and not tofu ?
Are the main & fork still compliant enough so it will work ?

eloyred:
Good question, and it made me change the default: `TERRAFORM_TOFU_VALIDATE` now
analyzes **`.tofu` only**, exactly like `TERRAFORM_TOFU_FMT`. So out of the box it
never touches a `.tf` file, and a Terraform repository is not affected at all.
`.tf` stays free for a future `TERRAFORM_TERRAFORM_VALIDATE`.

For users who do opt in with
`TERRAFORM_TOFU_VALIDATE_FILE_EXTENSIONS: [".tofu", ".tf"]`, I tested
Terraform-authored configuration against the real binary (`tofu` 1.12.6 from the
image) and could not produce a single false error. Each case is a directory run
through `tofu -chdir=<dir> init -backend=false -input=false -no-color` then
`tofu -chdir=<dir> validate -no-color`:

| Construct | Result |
|---|---|
| Plain HCL: `terraform {}` block, variables, locals, outputs | valid |
| `required_providers` with `hashicorp/aws` | resolved from the OpenTofu registry |
| `required_version = "~> 1.13"`, newer than OpenTofu's own 1.12.6 | valid |
| `required_version = ">= 99.0.0"` | valid — not enforced by `init -backend=false` nor by `validate` |
| `ephemeral = true` on a variable and on an output | supported |
| `ephemeral "random_password" "example" {}` resource block | supported |
| `terraform.applying` | supported |

The case I expected to bite was `required_version`: OpenTofu's version line sits
behind Terraform's, so a module pinning a newer core version looked like a
guaranteed false error. It is not — that check only happens on plan/apply. I also
hit it for real: a repository pinning `required_version = "1.11.8"`, an exact pin
the 1.12.6 binary does not satisfy, initialized and validated with no complaint.

What I can not prove is the absence of divergence for everything, in particular
provider-schema-driven features (write-only attributes, provider-defined
functions), since exercising those needs a real provider download. With `.tofu` as
the only default, that residual risk only reaches users who explicitly opt in.

Worth noting the mirror direction, which is the strongest argument for keeping the
extensions separate: OpenTofu-only syntax validates fine here and `terraform
validate` would reject it. I verified that both the `encryption` block and
`provider "null" { for_each = ... }` pass. So when a `TERRAFORM_TERRAFORM_VALIDATE`
arrives, each linter keeps its own dialect and neither reports false errors on the
other's.

nvuillam (approved):
Nice first PR, thanks :)

PR #8796, opened 2026-08-25
Keep a green run green when git cannot list updated files

### LINKED ISSUE #8649: Bug: UpdatedSourcesReporter can fail a passing run on a read-only workspace with git-lfs
Seen on MegaLinter 9.6.0 at [`5bb944a`][commit].

## Describe the bug

----
> 🦾 Written with LLM assistance [claude-opus-5]
> 💪 Reviewed by a human before submission
----

A run in which every linter passes exits non-zero. The exit happens after the SARIF report has been written, because a reporting helper raises an unhandled `git.exc.GitCommandError`.

[`UpdatedSourcesReporter.produce_report()`][usr-produce-report] calls [`utils.list_updated_files()`][utils-list-updated-files], which ends on an unguarded `repo.index.diff(None)` ([`utils.py:409`][utils-diff-call]). GitPython runs that as the `cmdline` shown in the traceback below.

`list_updated_files()` catches `git.InvalidGitRepositoryError` in two places, but not `GitCommandError`. The error travels up through [`MegaLinter.run()`][ml-report-loop] and ends the process:

```text
File "/megalinter/MegaLinter.py", line 383, in run
  reporter.produce_report()
File "/megalinter/reporters/UpdatedSourcesReporter.py", line 36, in produce_report
  updated_files = utils.list_updated_files(self.master.github_workspace)
File "/megalinter/utils.py", line 372, in list_updated_files
  changed_files = [item.a_path for item in repo.index.diff(None)]
...
git.exc.GitCommandError: Cmd('git') failed due to: exit code(128)
  cmdline: git diff --abbrev=40 --full-index -M --raw -z --no-color
```

This took a while to pin down because GitPython's `handle_process_output` drains the child's stderr before `GitCommandError` is constructed. The exception carries `stderr: b''`, so git's own message never reaches the log. The same empty stderr appears in [#1618][issue-1618].

<details>
<summary>Where the guard is, and where it isn't</summary>

The per-linter CONSOLE reporter loop treats a reporter failure as non-fatal ([`MegaLinter.py:360-369`][ml-console-loop]):

```python
if reporter.name == "CONSOLE" and reporter.is_active:
    try:
        reporter.produce_report()
    except Exception as e:
        logging.error("Unable to process reporter " + reporter.name + str(e))
```

The mega-linter-scope loop that follows doesn't have the same wrapper ([`MegaLinter.py:386-388`][ml-report-loop]):

```python
# Generate reports
for reporter in self.reporters:
    reporter.produce_report()
```

</details>

Because [`SarifReporter` sets `processing_order = -9999  # Run first`][sarif-order], the SARIF file is complete and on disk by the time the crash happens. A consumer ends up with a valid SARIF reporting no findings, plus a non-zero exit code.

## Root cause

Three conditions have to hold at once:

1. The workspace is bind-mounted read-only. That's a common hardening choice when `APPLY_FIXES: none`, since MegaLinter has no reason to write to the sources.

2. The repo has a required content filter. A `.gitattributes` entry such as `docs/demo/*.gif filter=lfs` qualifies, because the MegaLinter image ships `git-lfs` ([`Dockerfile:466`][dockerfile-lfs], [`Dockerfile-custom-flavor:20`][dockerfile-custom-flavor-lfs]) and the Alpine package runs `git lfs install --system` on install.

3. At least one index entry is stat-dirty, which makes git re-read work-tree content instead of trusting the cached stat. The usual cause is a uid difference. The index gets written by the CI runner or host user, the diff runs as container root, and every entry then mismatches. Anything that perturbs ctime, ino or mtime does the same, including SELinux relabeling on a bind mount or a fresh checkout.

<details>
<summary>Checking condition 2 inside alpine:3.24</summary>

```console
$ podman run --rm alpine:3.24 sh -c 'apk add -q git git-lfs; cat /etc/gitconfig'
[filter "lfs"]
        clean = git-lfs clean -- %f
        smudge = git-lfs smudge -- %f
        process = git-lfs filter-process
        required = true
```

</details>

Re-reading a `filter=lfs` file runs the LFS clean filter, which needs scratch space under `.git/lfs/tmp`. On a read-only mount it can't create it:

```text
Error cleaning Git LFS object: open /tmp/lint/.git/lfs/tmp/1661369123: no such file or directory
error: external filter 'git-lfs filter-process' failed
fatal: docs/demo/tour.gif: clean filter 'lfs' failed
```

The LFS pointer files are unmodified. Once git can run the filter, the diff comes back empty. Nothing here is a real content change; the filter simply has nowhere to write.

`git status` won't show this, since it refreshes the index and has cheaper paths available on a clean tree. `repo.index.diff(None)`, which is what `list_updated_files()` calls, goes straight to the stat comparison and the content re-read.

<details>
<summary>Why the reporter is the only caller that reaches this under APPLY_FIXES: none</summary>

The other caller, [`Linter.update_files_lint_results()`][linter-try-fix], is gated on `self.try_fix`:

```python
if self.try_fix is True:
    updated_files = utils.list_updated_files(self.github_workspace)
else:
    updated_files = []
```

With fixes disabled, `UpdatedSourcesReporter` is the only entry point left. It is [active by default][usr-activation], and `manage_activation()` consults only `can_write_report_files()` and `UPDATED_SOURCES_REPORTER`.

</details>

## To Reproduce

MegaLinter isn't needed to trigger the underlying git failure. The command below is the one GitPython issues.

```bash
# Isolate from the caller's git identity/config
export GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null
export GIT_AUTHOR_NAME=repro GIT_AUTHOR_EMAIL=repro@invalid
export GIT_COMMITTER_NAME=repro GIT_COMMITTER_EMAIL=repro@invalid

git init -q repo && cd repo && mkdir -p docs/demo
printf 'docs/demo/*.gif filter=lfs diff=lfs merge=lfs -text\n' > .gitattributes
printf 'version https://git-lfs.github.com/spec/v1\noid sha256:%064d\nsize 12345\n' 1 > docs/demo/tour.gif
git add .gitattributes docs/demo/tour.gif && git commit -q -m init && cd ..

# Read-only workspace, as MegaLinter sees it on a hardened mount
podman run --rm -v "$PWD/repo:/tmp/lint:ro,z" alpine:3.24 sh -c \
  'apk add -q git git-lfs && git config --global --add safe.directory /tmp/lint &&
   cd /tmp/lint && git diff --abbrev=40 --full-index -M --raw -z --no-color; echo exit=$?'
# exit=128
```

One thing to watch for: run the read-only case first. Any run with a writable `.git` refreshes the on-disk index to the container's uid, which leaves every later read-only run stat-clean and hides the problem. It's worth recreating the fixture between attempts.

<details>
<summary>Verification matrix: four single-variable runs</summary>

Each row changes one variable from the failing case. Run on `alpine:3.24` with git-lfs 3.7.1, the same Alpine release the MegaLinter runtime stage builds on ([`Dockerfile:153`][dockerfile-base-image]: `FROM python:3.14-alpine3.24`).

| # | Workspace           | `git-lfs`     | Index uid vs container uid | `git diff-files` (stat-dirty?) | Result                                   |
| - | ------------------- | ------------- | -------------------------- | ------------------------------ | ---------------------------------------- |
| 1 | `:ro`               | installed     | 1000 vs 0, mismatch        | both files listed              | `exit=128`, clean filter failed          |
| 2 | `:ro`, `.git` `:rw` | installed     | 1000 vs 0, mismatch        | both files listed              | `exit=0` (and the index gets refreshed)  |
| 3 | `:ro`               | not installed | 1000 vs 0, mismatch        | both files listed              | `exit=0`, an undefined filter is a no-op |
| 4 | `:ro`               | installed     | 0 vs 0, match              | nothing listed                 | `exit=0`, stat trusted, no re-read       |

Row 4 is the evidence for condition 3. With an index whose recorded uid matches the container, the same command returns 0 on the same read-only mount with the same LFS filter configured. So does every flag subset of it: `git diff`, `--raw`, `-M --raw`, and the full GitPython line.

> Note: The git failure was reproduced directly rather than through a MegaLinter image.

</details>

## Expected behavior

A run where every linter passes exits 0. A reporting helper that can't determine which files changed can report no updated files and log a warning.

`list_updated_files()` already treats "cannot determine updated files" as non-fatal in two places. It returns `[]` with a `logging.warning` for a missing repository, and again for a non-root working copy ([`utils.py:395-408`][utils-warning-returns]). The diff on the final line is the one case without that treatment.

### 1. Catch git.GitCommandError in list_updated_files()

Return `[]` and log a warning naming the failed command and the workspace, so the condition stays visible. That covers all three callers of the helper.

There's a judgement call here worth flagging. The same helper feeds the fixed-file counter through [`Linter.py:1067-1068`][linter-try-fix] and `utils.check_updated_file()`.

Catching inside the helper fixes every caller at once, but under `APPLY_FIXES: all` it converts a hard failure into a reported "0 files fixed". Catching in the reporter instead keeps that path loud, and leaves `Linter.py:1068` exposed to the same crash. Both look defensible from here; the warning text is what makes the first option safe.

Because the exception's `stderr` is empty, that warning needs to carry the command and the workspace path itself. Otherwise the user sees "unable to list updated files" with no way to find out why.

### 2. Deactivate UpdatedSourcesReporter when APPLY_FIXES is none

`manage_activation()` already has the hook for this. One caveat, from [`Linter.py:1525-1526`][linter-fix-flag]:

```python
# Some linters/formatters update files by default.
# To avoid that, declare -megalinter-fix-flag as cli_lint_fix_arg_name
```

So `APPLY_FIXES: none` doesn't guarantee an unmodified workspace. A linter whose descriptor omits that declaration still rewrites files, and this reporter is what surfaces it. Turning the reporter off unconditionally would drop that signal. Fix 1 stands on its own if you'd rather keep the change small.

## Test coverage

Nothing under `megalinter/tests/` references either symbol today:

```console
$ grep -rl "list_updated_files\|UpdatedSourcesReporter" megalinter/tests/
$ # (no matches)
```

A unit test that injects a `GitCommandError` from `repo.index.diff` and asserts `list_updated_files()` returns `[]` would pin fix 1 down without needing a container, a read-only mount or git-lfs.

## Impact

Every full-codebase run fails, deterministically, for any repo that combines a read-only workspace with git-lfs. The crash is in reporting rather than linting, so the exit code contradicts a green, fully written SARIF. The empty `stderr` makes the cause hard to identify from the logs alone.

## Workaround

[`UPDATED_SOURCES_REPORTER: false`][usr-docs] avoids it today. Bind-mounting `.git` writable also works (row 2 of the matrix), though that lets git refresh the index, which then hides the problem on later runs.

<details>
<summary>Earlier reports touching the same helper</summary>

Three earlier issues involve `list_updated_files()`, arriving through two different callers. #1618 and #1882 came in via the linter caller (`Linter.update_files_lint_results`); #4270 and this report come in via `UpdatedSourcesReporter.produce_report()`. Each was fixed where it surfaced, so the helper's own error handling has stayed as it is.

- [#1882, _Fatal runtime error occurs when using Git LFS_][issue-1882] (closed 2022-09-19). Same helper, same `exit code(128)`, also git-lfs, but the opposite root cause: git-lfs was missing from the image. [PR #1883][pr-1883] added it. That package is the one whose required clean filter is involved here, which is why the two reports are linked.
- [#1618, _`git.exc.GitCommandError: Cmd('git') failed due to: exit code(128)`_][issue-1618] (closed 2022-09-05). Same traceback out of `list_updated_files`, on Azure DevOps. The investigation stalled on the empty `stderr`, and the issue was later closed as stale.
- [#4270, _\[8.2.0\] Updated Sources Reporter issue_][issue-4270] (closed 2024-11-18). The closest precedent: an `AttributeError` in this same `produce_report()`, reported as "Exit code 1, so all runs fail even though all linters are green."
- [#6455][issue-6455] and [PR #6456][pr-6456], a `GitCommandError` from git worktrees. The PR proposed catching `GitCommandError` around git operations and continuing with a warning, and was closed unmerged. The approach overlaps with fix 1 above.
- [#2473, _CSpellLinter / SPELL_CSPELL fails to run in a read-only filesystem_][issue-2473] (closed 2023-03-25), which shows read-only workspaces are a shape people run.

I couldn't find an open issue covering this one, so I don't think it's a duplicate, though I may have missed something.

</details>

<details>
<summary>Additional context</summary>

- MegaLinter sets `safe.directory` for the workspace in [`entrypoint.sh:32-33`][entrypoint-safe-directory], which the reproduction mirrors.
- GitPython's stderr handling lives in [`git/cmd.py`][gitpython-cmd], in `handle_process_output` and `finalize_process`. The stream is drained by the output handler before `GitCommandError` is raised, which is why the exception's `stderr` is empty.
- Required filters are documented under `filter` in [gitattributes(5)][gitattributes]. A `required` filter turns a filter failure into a fatal error rather than a pass-through.

</details>

[commit]: https://github.com/oxsecurity/megalinter/commit/5bb944a715b995cb0dd49672b2a1afb76e3c12a6
[usr-produce-report]: https://github.com/oxsecurity/megalinter/blob/5bb944a715b995cb0dd49672b2a1afb76e3c12a6/megalinter/reporters/UpdatedSourcesReporter.py#L36
[usr-activation]: https://github.com/oxsecurity/megalinter/blob/5bb944a715b995cb0dd49672b2a1afb76e3c12a6/megalinter/reporters/UpdatedSourcesReporter.py#L19-L31
[utils-list-updated-files]: https://github.com/oxsecurity/megalinter/blob/5bb944a715b995cb0dd49672b2a1afb76e3c12a6/megalinter/utils.py#L395-L410
[utils-warning-returns]: https://github.com/oxsecurity/megalinter/blob/5bb944a715b995cb0dd49672b2a1afb76e3c12a6/megalinter/utils.py#L395-L408
[utils-diff-call]: https://github.com/oxsecurity/megalinter/blob/5bb944a715b995cb0dd49672b2a1afb76e3c12a6/megalinter/utils.py#L409
[ml-report-loop]: https://github.com/oxsecurity/megalinter/blob/5bb944a715b995cb0dd49672b2a1afb76e3c12a6/megalinter/MegaLinter.py#L386-L388
[ml-console-loop]: https://github.com/oxsecurity/megalinter/blob/5bb944a715b995cb0dd49672b2a1afb76e3c12a6/megalinter/MegaLinter.py#L360-L369
[sarif-order]: https://github.com/oxsecurity/megalinter/blob/5bb944a715b995cb0dd49672b2a1afb76e3c12a6/megalinter/reporters/SarifReporter.py#L29
[linter-try-fix]: https://github.com/oxsecurity/megalinter/blob/5bb944a715b995cb0dd49672b2a1afb76e3c12a6/megalinter/Linter.py#L1067-L1068
[linter-fix-flag]: https://github.com/oxsecurity/megalinter/blob/5bb944a715b995cb0dd49672b2a1afb76e3c12a6/megalinter/Linter.py#L1525-L1526
[dockerfile-lfs]: https://github.com/oxsecurity/megalinter/blob/5bb944a715b995cb0dd49672b2a1afb76e3c12a6/Dockerfile#L466
[dockerfile-base-image]: https://github.com/oxsecurity/megalinter/blob/5bb944a715b995cb0dd49672b2a1afb76e3c12a6/Dockerfile#L153
[dockerfile-custom-flavor-lfs]: https://github.com/oxsecurity/megalinter/blob/5bb944a715b995cb0dd49672b2a1afb76e3c12a6/Dockerfile-custom-flavor#L20
[entrypoint-safe-directory]: https://github.com/oxsecurity/megalinter/blob/5bb944a715b995cb0dd49672b2a1afb76e3c12a6/entrypoint.sh#L32-L33

[issue-1618]: https://github.com/oxsecurity/megalinter/issues/1618
[issue-1882]: https://github.com/oxsecurity/megalinter/issues/1882
[pr-1883]: https://github.com/oxsecurity/megalinter/pull/1883
[issue-2473]: https://github.com/oxsecurity/megalinter/issues/2473
[issue-4270]: https://github.com/oxsecurity/megalinter/issues/4270
[issue-6455]: https://github.com/oxsecurity/megalinter/issues/6455
[pr-6456]: https://github.com/oxsecurity/megalinter/pull/6456

[usr-docs]: https://megalinter.io/latest/reporters/UpdatedSourcesReporter/
[gitpython-cmd]: https://github.com/gitpython-developers/GitPython/blob/main/git/cmd.py
[gitattributes]: https://git-scm.com/docs/gitattributes

trevor-vaughan:
----
> 🦾 Written with LLM assistance [claude-opus-5]
> 💪 Reviewed by a human before submission
----

## Summary

A run where every linter passed could still exit with an error. `list_updated_files()` ran `repo.index.diff(None)` without a guard, and on a read-only workspace whose repository uses git-lfs the clean filter has nowhere to write its temporary files, so the diff exits 128 and `UpdatedSourcesReporter.produce_report()` propagated a `GitCommandError`.

Listing updated files is best effort — the helper already returned `[]` with a warning for the two other cases it cannot answer, so this adds `GitCommandError` to that set. The `UPDATED_SOURCES_REPORTER: false` workaround is no longer needed.

## Changes

- `megalinter/utils.py`: catch `git.GitCommandError` in `list_updated_files()` and return `[]`. GitPython drains stderr before raising, so the warning carries the workspace path and the failed command instead — that is all the user gets to diagnose from
- Warn once per workspace: `Linter.update_files_lint_results` calls the helper once per linted file in `file` mode, so subsequent failures for the same workspace log at debug level
- `megalinter/tests/test_megalinter/utils_test.py`: tests for the nominal listing, the git-lfs failure, the warn-once behavior, and both non-repository cases. Git config is neutralized (`GIT_CONFIG_GLOBAL`, `GIT_CONFIG_SYSTEM`, author/committer) so the caller's own filters never influence the diff
- `docs/reporters/UpdatedSourcesReporter.md`: new "Empty updated sources folder" section explaining the cause and the `.git`-writable fix
- `CHANGELOG.md` entry under Core

Fixes #8649

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,9 @@ Note: Can be used with `oxsecurity/megalinter@beta` in your GitHub Action mega-l
     - **REPOSITORY_CHECKOV** and **REPOSITORY_BETTERLEAKS** only analyze the Pull Request changes when asked to
     - Set `BITBUCKET_PR_ID` in your pipeline (Bitbucket provides it on Pull Request builds) to benefit from it
     - Fixed JSON config schema for Betterleaks
+  - A run where all linters pass does not **exit with an error** anymore when MegaLinter can not list the files updated by the linters ([#8649](https://github.com/oxsecurity/megalinter/issues/8649))
+    - Happens on a **read-only workspace** whose repository uses **git-lfs**: the required LFS filter has nowhere to write its temporary files, so the `git diff` used to detect updated files exits 128
+    - MegaLinter now logs a **warning** naming the workspace and the failed command, reports no updated source file, and completes the run. The `UPDATED_SOURCES_REPORTER: false` workaround is not needed anymore
 
 - Reporters
   - Linters reporting in **SARIF** format no longer show **No output available** in Pull Request comments and summaries: the details section now names the SARIF report to open and links the **MegaLinter artifacts** ([#8730](https://github.com/oxsecurity/megalinter/issues/8730))
--- a/docs/reporters/UpdatedSourcesReporter.md
+++ b/docs/reporters/UpdatedSourcesReporter.md
@@ -34,6 +34,12 @@ If you aren't using GitHub Actions, you can:
 - use [Email Reporter](EmailReporter.md): Updated source folder will be in the email attachment reports zip
 - publish folder `<WORKSPACE>/report/updated_sources` as artifact with your CI tool
 
+### Empty updated sources folder
+
+The folder is empty when MegaLinter can not list the files updated by the linters. This happens on a **read-only workspace** whose repository uses **git-lfs**: the LFS filter has nowhere to write its temporary files, so the `git diff` used to detect updated files fails.
+
+MegaLinter logs a warning naming the workspace and the failed command, then completes the run. Mount `.git` as writable to get the updated sources back.
+
 ## Configuration
 
 | Variable                     | Description                                             | Default value   |
--- a/megalinter/tests/test_megalinter/utils_test.py
+++ b/megalinter/tests/test_megalinter/utils_test.py
@@ -4,14 +4,23 @@ Unit tests for utils class
 
 """
 
+import os
 import re
+import subprocess
+import tempfile
 import unittest
 import uuid
 import warnings
+from unittest.mock import MagicMock, patch
 
+import git
 from megalinter import config
 from megalinter.logger import fetch_betterleaks_regexes, sanitize_string
-from megalinter.utils import fix_regex_pattern, get_excluded_directories
+from megalinter.utils import (
+    fix_regex_pattern,
+    get_excluded_directories,
+    list_updated_files,
+)
 
 
 class utils_test(unittest.TestCase):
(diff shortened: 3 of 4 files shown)


nvuillam (approved):
Good PR, thanks @trevor-vaughan 😊

nvuillam:
@trevor-vaughan plz can u merge conflicts so i can merge ? :)

trevor-vaughan:
@nvuillam Done

nvuillam:
Thanks:)
I'll generate a new beta today so u can test on your project:)

